PR #13, opened 2024-07-10
Revit 2022 for DUI3

adamhathcock:
2022 loads

https://spockle.atlassian.net/browse/DUI3-395


Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/SelectionBinding.cs
```diff
@@ -26,8 +26,10 @@ ITopLevelExceptionHandler topLevelExceptionHandler
     _topLevelExceptionHandler = topLevelExceptionHandler;
     // POC: we can inject the solution here
     // TODO: Need to figure it out equivalent of SelectionChanged for Revit2020
+#if !REVIT2020 && !REVIT2021 && !REVIT2022
```
BovineOx, line 29: We were talking about being able to say - is this 2022 onwards for instance, to simplify this

Converters/Revit/Speckle.Converters.Revit2022.DependencyInjection/RevitConverterModule.cs
BovineOx, line 1: We need separate files/projects? I guess it's not a huge amount but I worry if DI is forgotten across versions that would be common to all

DUI3/Speckle.Connectors.DUI/Url.cs
```diff
@@ -12,7 +12,8 @@
 /// </remarks>
 public static class Url
 {
-  public static readonly Uri Netlify = new("https://boisterous-douhua-e3cefb.netlify.app/");
+  //old beta
+  public static readonly Uri Netlify = new("https://6671ddd20235b90008dbc453--boisterous-douhua-e3cefb.netlify.app/");
```
BovineOx, line 16: Is this no longer in he XAML at all? We will need to be careful not to accidentally take this forever. Does @oguzhankoral have  aticket now for fixing the CEF 60 something support? If so can we have a subtask and drop it here to update the URL back to the right one plz?

BovineOx (changes requested):
Not much really, request to not lose sight of changing the URL back and mostly questions around whether we can have pre-processor directives of REVIT_2022_ONWARDS or something, I think this was the previous suggestion but a) not mandatory, just maybe the right time to do it and b) @AlanRynne should weigh-in

I just stops having to faff and extra lines in places, but the !<this version> && !<that version> clearly works

Converters/Revit/Speckle.Converters.Revit2022.DependencyInjection/RevitConverterModule.cs
adamhathcock: We have to have new projects and refs....the DI project is not really used (I'd like to remove it and the concept) but is needed for now...was working on a shared project for it to keep everything in sync

BovineOx: The DI projects were an effort to keep DI concepts outside of things like conversion so we're aren't pushing Autofac on people, so they probably should continue to exist

Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/SelectionBinding.cs
AlanRynne: We do have this in RHINO, following the standard pattern used for c# frameworks: `RHINO7_OR_GREATER`.

I really suggest we do this bc it allows for setting lower/upper bounds without having to explicitly state all versions.

Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/SelectionBinding.cs
```diff
@@ -25,9 +25,11 @@ ITopLevelExceptionHandler topLevelExceptionHandler
     _revitIdleManager = idleManager;
     _topLevelExceptionHandler = topLevelExceptionHandler;
     // POC: we can inject the solution here
-    // TODO: Need to figure it out equivalent of SelectionChanged for Revit2020
+    // TODO: Need to figure it out equivalent of SelectionChanged for Revit2022
```
BovineOx, line 28: remove POC: add ticket to DUI3 project and link here for investigate solution for 2022

Connectors/Revit/Speckle.Connectors.RevitShared/DependencyInjection/RevitConnectorModule.cs
```diff
@@ -73,6 +73,11 @@ public void RegisterUiDependencies(SpeckleContainerBuilder builder)
     builder.AddSingleton<IRevitPlugin, RevitWebViewPlugin>();
     builder.AddSingleton<DUI3ControlWebView>();
     builder.AddSingleton<DUI3ControlWebViewDockable>();
+#elif REVIT2022
+    // POC: different versons for different versions of CEF
```
BovineOx, line 77: if this is just as is, we can remove the POC: portion of comment

BovineOx (approved):

BovineOx:
Just add the ticket URL for selection change

Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/SelectionBinding.cs
adamhathcock: https://spockle.atlassian.net/browse/CNX-9874

PR #20, opened 2024-07-11
DUI3-350 update dui3 connectors to use new api client functions

JR-Morgan:
Updated DUI3 to call new Client function
This meant changing the ReceoveInfo to take a few extra props (the modelId and the sourceApplication slug)

### DIFF
--- a/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitReceiveBinding.cs
+++ b/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitReceiveBinding.cs
@@ -3,6 +3,7 @@ using Speckle.Connectors.DUI.Bindings;
 using Speckle.Connectors.DUI.Bridge;
 using Speckle.Connectors.DUI.Models;
 using Speckle.Connectors.DUI.Models.Card;
+using Speckle.Connectors.Revit.Plugin;
 using Speckle.Connectors.Utils;
 using Speckle.Connectors.Utils.Builders;
 using Speckle.Connectors.Utils.Cancellation;
@@ -16,6 +17,7 @@ internal class RevitReceiveBinding : IReceiveBinding
   public string Name => "receiveBinding";
   public IBridge Parent { get; }
 
+  private readonly RevitSettings _revitSettings;
   private readonly CancellationManager _cancellationManager;
   private readonly DocumentModelStore _store;
   private readonly IUnitOfWorkFactory _unitOfWorkFactory;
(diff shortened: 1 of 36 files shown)


Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitReceiveBinding.cs
```diff
@@ -55,7 +59,7 @@ public async Task Receive(string modelCardId)
       // Receive host objects
       HostObjectBuilderResult conversionResults = await unitOfWork
         .Service.Execute(
-          modelCard.GetReceiveInfo(),
+          modelCard.GetReceiveInfo(_revitSettings.HostSlug.NotNull()),
```
AlanRynne, line 62: Is there a chance the hostSlug would not be null? 🤔

DUI3/Speckle.Connectors.DUI/Models/Card/ReceiverModelCard.cs
```diff
@@ -12,13 +12,15 @@ public class ReceiverModelCard : ModelCard
...
     new(
       AccountId.NotNull(),
       new Uri(ServerUrl.NotNull()),
       ProjectId.NotNull(),
       ProjectName.NotNull(),
+      ModelId.NotNull(),
       ModelName.NotNull(),
-      SelectedVersionId.NotNull()
+      SelectedVersionId.NotNull(),
+      sourceApplication
```
AlanRynne, line 24: I'm assuming you already looked into injecting the source application somehow but this was the easiest way through?

It seems a bit odd to have to pass in the hostApp to get the "receive info"

AlanRynne (commented):
Looks good, just 2 comments

DUI3/Speckle.Connectors.DUI/Models/Card/ReceiverModelCard.cs
JR-Morgan: It is, I agree. but this is what we are doing with the SendInfo.

I think we should proably come up with a different design. Our settings (`RhinoSettings`, `AutoCADSettings`, etc..) are all pretty different looking, which is a shame. I might ping @BovineOx see if we should create a ticket

adamhathcock (approved):

Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitReceiveBinding.cs
JR-Morgan: No it should never be null, you're right.
`HostSlug` is not nullable... I just copied this block from the send bindings.

However, likley we should refactor these settings to share a common interface, since right now all connector settings look wildly different from one another despite basically just storing the host app version slugs.

JR-Morgan: https://spockle.atlassian.net/browse/DUI3-484

DUI3/Speckle.Connectors.DUI/Models/Card/ReceiverModelCard.cs
JR-Morgan: https://spockle.atlassian.net/browse/DUI3-484

PR #41, opened 2024-07-17
More warnings

JR-Morgan:
1. Added back the analyser settings that we had in the old sharp repo, with some additional tweaks for new C#12 analysers.
2. Fixed all the warnings that were introduced during the time that we had these analysers disabled
    - the only ones that may raise some eyebrows are Revit element disposal
4. Fixed some of the new analysers


Converters/Revit/Speckle.Converters.RevitShared/ToHost/TopLevel/ToDirectShape/CurveToDirectShapeConverterToHostBase.cs
```diff
@@ -33,7 +33,7 @@ public List<DB.GeometryObject> Convert(TCurve target)
       DB.BuiltInCategory.OST_GenericModel
     );
 
-    var revitDs = DB.DirectShape.CreateElement(_contextStack.Current.Document, genericModelCategory.Id);
+    using var revitDs = DB.DirectShape.CreateElement(_contextStack.Current.Document, genericModelCategory.Id);
```
JR-Morgan, line 36: @connorivy in a few places where we are creating revit elements, but they are falling out of scope (e.g. we are not returning them, or passing them into another object),
I'm adding a `using` keyword to dispose the C# object.

In my testing, this appears to work fine, the object still converts as expected. Adding the `using` keywords might help us be deterministic with disposal, and perhaps lower our memory footprint.

The alternative is we start silencing CA2000 just for Revit, but I don't see any downsides to what I'm doing here.

Directory.Build.props
```diff
@@ -1,15 +1,52 @@
...
+      CA1303;CA1304;CA1305;CA1307;CA1308;CA1309;CA1310;CA1311;
+      <!-- Logging -->
+      CA1848;CA2254;CA1727;
+      <!-- Others we don't want -->
+      CA1815;CA1725;
+      $(NoWarn)
+    </NoWarn>
+  </PropertyGroup>
+
+  <PropertyGroup Label="Repo Info">
```
JR-Morgan, line 37: @adamhathcock did you add this somewhere else? or was that just for our core nugets?

Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/SelectionBinding.cs
```diff
@@ -4,13 +4,16 @@
 using Speckle.Connectors.Revit.Plugin;
 using Speckle.Converters.Common;
 using Speckle.Converters.RevitShared.Helpers;
-using Timer = System.Timers.Timer;
 
 namespace Speckle.Connectors.Revit.Bindings;
 
 // POC: we need a base a RevitBaseBinding
-internal sealed class SelectionBinding : RevitBaseBinding, ISelectionBinding
+internal sealed class SelectionBinding : RevitBaseBinding, ISelectionBinding, IDisposable
```
JR-Morgan, line 11: @adamhathcock I'm hoping (but not tested) that autofac is smart enough to dispose bindings...

adamhathcock (approved):

Converters/Revit/Speckle.Converters.RevitShared/ToHost/TopLevel/ToDirectShape/CurveToDirectShapeConverterToHostBase.cs
connorivy: I'm not aware of any reason this wouldn't be okay

connorivy (approved):

PR #63, opened 2024-07-23
feat(rhino): DUI3-142 adds render materials to objects on send and receive

clairekuang:
Adds render materials to atomic objects on send and receive. 
Behavior notes:
- rhino objects and layers have a `MaterialSource` attribute, which by default is set to `ByLayer`. This pr sends a `renderMaterialId` on every object, even if it is inheriting its material. This means that on receive, the `MaterialSource` property is not preserved
- the `MaterialSource` property on receive is set to `ByObject` if a valid `renderMaterialId` is found on the object, or `ByLayer` if no material is found (default material is used in this case)
- materials are cleansed from the doc on receive by the `baseLayerName` (same as for block definitions). They are named with the same convention as block definitions.

TODO: add render materials to layers

image


Connectors/Rhino/Speckle.Connectors.Rhino7/HostApp/RhinoMaterialManager.cs
```diff
@@ -0,0 +1,147 @@
...
+
+  /// <summary>
+  /// Creates a Speckle Render Material from the provided Rhino material
+  /// </summary>
+  /// <param name="material"></param>
+  /// <returns>The existing Speckle Render Material if this material has been previously created, or the newly created Speckle Render Material</returns>
+  public SpeckleRenderMaterial CreateSpeckleRenderMaterial(Material material)
+  {
+    string materialId = material.Id.ToString();
+    if (_renderMaterialCache.ContainsKey(materialId))
```
oguzhankoral, line 31: CI has already valid idea

Connectors/Rhino/Speckle.Connectors.Rhino7/HostApp/RhinoMaterialManager.cs
```diff
@@ -0,0 +1,147 @@
...
+    {
+      return _renderMaterialCache[materialId];
+    }
+
+    // get physically based render material
+    Material pbMaterial = material;
+    if (!material.IsPhysicallyBased)
```
oguzhankoral, line 38: I'm a noob on Rhino materials. Why do we need to check it like this?

oguzhankoral: Maybe you can show me later on Rhino

Connectors/Rhino/Speckle.Connectors.Rhino7/HostApp/RhinoMaterialManager.cs
```diff
@@ -0,0 +1,147 @@
...
+          {
+            Name = matName,
+            DiffuseColor = diffuse,
+            EmissionColor = emissive,
+            Transparency = transparency
+          };
+
+        int matIndex = doc.Materials.Add(rhinoMaterial);
+
+        // POC: check on matIndex -1, means we haven't created anything - this is most likely an recoverable error at this stage
```
oguzhankoral, line 110: or does it mean we have it already with the same name? This is also something we need to check if you haven't already. "What will happen if we have the materials with the same name already?"

Connectors/Rhino/Speckle.Connectors.Rhino7/HostApp/RhinoMaterialManager.cs
```diff
@@ -0,0 +1,147 @@
...
+  /// <summary>
+  /// Removes all materials with a name starting with <paramref name="namePrefix"/> from the active document
+  /// </summary>
+  /// <param name="namePrefix"></param>
+  public void PurgeMaterials(string namePrefix)
+  {
+    var currentDoc = RhinoDoc.ActiveDoc; // POC: too much right now to interface around
+    foreach (Material material in currentDoc.Materials)
+    {
+      if (!material.IsDeleted && material.Name.Contains(namePrefix))
```
oguzhankoral, line 141: No action needed: later we should handle these checks with UserDictionaries instead of messing with the name of materials.

oguzhankoral:
I guess we do not receive again if the material name is "default" which seems to me a bit fragile to have it with that simple naming. Maybe it is worth to consider add prop to render material object as `isDefault`? @clairekuang

Connectors/Rhino/Speckle.Connectors.Rhino7/HostApp/RhinoMaterialManager.cs
clairekuang: Any material with the same name should've been purged already

clairekuang:
> I guess we do not receive again if the material name is "default" which seems to me a bit fragile to have it with that simple naming. Maybe it is worth to consider add prop to render material object as `isDefault`? @clairekuang

not necessarily - depends on the expected behavior when receiving default materials in other applications. Eg the default material in rhino != default material in autocad, i'd rather receive the rhino default material in autocad rather than set incoming objs to the autocad default

oguzhankoral (approved):

PR #66, opened 2024-07-23
Autocad cleanup

adamhathcock:


oguzhankoral (approved):
Tested both acad and revit

PR #76, opened 2024-07-25
Top Level Exception Handler and Bindings fix

JR-Morgan:
With the grand help of @oguzhankoral, We've fixed the `TopLevelExceptionHandler` so that it can display global toast notifications for non-top level binding usages.

CLARIFICATIONS POST PR:
 - All bindings that want to use the `TopLevelExceptionHandler` now should NOT inject, instead they should use their injected `IBridge Parent` whom now has a `TopLevelExceptionHandler` property.
 - All non-binidngs (such as DocumentStores) that want to use a TLevelExeptHandler should inject an `ITopLevelExceptionHandler`, this one will be using internally a dummy `TopLevelExceptionHandlerBinding`
 - ALL `BrowserBridg`es should be associated with a binding before usage, we are NOT accepting `BrowserBridg`es  to be used without a binding.

CHANGES:
 - Added TopLevelExceptionHandlerBinding
 - ITopLevelExceptionHandler is now registered as a factory of ^ rather than as a singleton.
 - Refactored the bridge's AccociateWithBinding function to now only take in binding information, browser information is now injected through the `IBrowserScriptExecutor` interface
 - Fixed Idle manager to ensure removal even if the call action is exceptional.
- Exposed `Speckle.Autofac`'s dependency on `Autofac` so that we can use `Autofac` in the connector modules (for registration of the same service to two interfaces)
 
Feedback welcome on interface naming

---

@adamhathcock please can you review my changes to the test project (test + internals visible to)
@oguzhankoral please can you review the remainder of the changes.


DUI3/Speckle.Connectors.DUI/Bridge/BrowserBridge.cs
```diff
@@ -296,11 +290,11 @@ private void NotifyUIMethodCallResultReady(string requestId, string? serializedD
   /// <returns></returns>
   public string? GetCallResult(string requestId)
   {
-    bool isFound = _resultsStore.TryRemove(requestId, out string? res);
-    if (!isFound)
-    {
-      throw new ArgumentException($"No result for the given request id was found: {requestId}", nameof(requestId));
-    }
+    _ = _resultsStore.TryRemove(requestId, out string? res);
```
oguzhankoral, line 293: I don't remember why we removed `isFound` logic, I would take it back if there is a no reason

JR-Morgan: Good catch, This was not intended to be left in this PR (we removed it briefly for experimental testing of unknown bindings)
I'll make sure it gets back in

oguzhankoral (approved):
Amazing work Jedd. LGTM!

PR #80, opened 2024-07-26
AGIS render materials send & receive

KatKatKateryna:
Tested with: https://latest.speckle.systems/projects/126cd4b7bb/models/0a4181c73b@f75f25ce4e 
image


oguzhankoral:
@KatKatKateryna I was about to merge your PR but having errors on tests, I'm preventing to merge it now, leaving it to you to fix.

z1imVAr32H

oguzhankoral (changes requested):
Having errors on receive

KatKatKateryna:
Tested with https://latest.speckle.systems/projects/126cd4b7bb/models/0a4181c73b , some objects are not colored but I don't see their IDs in the RenderMaterial Objects either:
image
@oguzhankoral

clairekuang (changes requested):
In general, it looks like you're pulling object and layer colors from the `diffuse` property of render materials. Autocad, Rhino, and Sketchup are now sending `ColorProxies` - this seems more appropriate for your color props.

After looking through arcgis documentation, wouldn't `RenderMaterial` be converted to a native `BasicMaterial` object instead? [https://pro.arcgis.com/en/pro-app/latest/sdk/api-reference/topic27295.html](https://pro.arcgis.com/en/pro-app/latest/sdk/api-reference/topic27295.html)

clairekuang:
We've decided to change this pr with Color Proxies instead, the first of which is https://github.com/specklesystems/speckle-sharp-connectors/pull/135 (on send)

clairekuang:
The rest of this pr can be closed without merge once Color Proxy receive is added as well

PR #83, opened 2024-07-29
receive lines and points as Instances

KatKatKateryna:
Tested with: https://latest.speckle.systems/projects/126cd4b7bb/models/6b62c61a22@c42b129063 
image


KatKatKateryna:
@oguzhankoral @clairekuang moved this to Transform function, with serialization hack as a backup for displayValues

Converters/ArcGIS/Speckle.Converters.ArcGIS3/Utils/LocalToGlobalConverterUtils.cs
```diff
@@ -30,100 +31,82 @@ public Base TransformObjects(Base atomicObject, List<Matrix4x4> matrix)
 
     // This is a temp hack. We would add transformations to conversions later instead try to copy objects like this.
```
BovineOx, line 32: If we consider this a hack we should add a linear ticket to unhack it otherwise it will stay a hack

Converters/ArcGIS/Speckle.Converters.ArcGIS3/Utils/LocalToGlobalConverterUtils.cs
```diff
@@ -30,100 +31,82 @@ public Base TransformObjects(Base atomicObject, List<Matrix4x4> matrix)
...
     // Jedd also has opinions on this.
-    Base newObject = Core.Api.Operations.Deserialize(Core.Api.Operations.Serialize(atomicObject));
+    // Base newObject = Core.Api.Operations.Deserialize(Core.Api.Operations.Serialize(atomicObject));
```
BovineOx, line 34: Why is this line commented out? We should remove if not needed now

Converters/ArcGIS/Speckle.Converters.ArcGIS3/Utils/LocalToGlobalConverterUtils.cs
```diff
@@ -30,100 +31,82 @@ public Base TransformObjects(Base atomicObject, List<Matrix4x4> matrix)
...
       {
-        return TransformObjWithDisplayValues(newObject, matrix);
+        c.TransformTo(transform, out ITransformable newObj);
+        c = newObj;
       }
-      throw new SpeckleException(
-        $"Display Values of types '{listVals.ToList().FindAll(x => x is not SOG.Mesh).Select(y => y.speckle_type).Distinct().ToList()}' for {newObject.speckle_type} are not supported for local to global coordinate transformation"
-      );
+
+      atomicObject = (Base)c;
```
BovineOx, line 47: This will throw if c cannot cast to Base, is this the intention here?

Converters/ArcGIS/Speckle.Converters.ArcGIS3/Utils/LocalToGlobalConverterUtils.cs
```diff
@@ -30,100 +31,82 @@ public Base TransformObjects(Base atomicObject, List<Matrix4x4> matrix)
...
-    if (newObject is SOG.Mesh mesh)
+
+    if (atomicObject.TryGetDisplayValue() is IReadOnlyList<Base> listVals)
     {
-      return TransformMesh(mesh, matrix);
+      if (listVals.ToList().FindAll(x => x is not ITransformable).Count == 0)
+      {
+        return TransformObjWithDisplayValues(atomicObject, transforms);
+      }
+      throw new SpeckleConversionException(
```
BovineOx, line 62: Whitespace should come after closing brace for readability

Converters/ArcGIS/Speckle.Converters.ArcGIS3/Utils/LocalToGlobalConverterUtils.cs
```diff
@@ -30,100 +31,82 @@ public Base TransformObjects(Base atomicObject, List<Matrix4x4> matrix)
...
-        return TransformObjWithDisplayValues(newObject, matrix);
+        c.TransformTo(transform, out ITransformable newObj);
+        c = newObj;
       }
-      throw new SpeckleException(
-        $"Display Values of types '{listVals.ToList().FindAll(x => x is not SOG.Mesh).Select(y => y.speckle_type).Distinct().ToList()}' for {newObject.speckle_type} are not supported for local to global coordinate transformation"
-      );
+
+      atomicObject = (Base)c;
+      foreach (var prop in atomicObject.GetMembers(DynamicBaseMemberType.Dynamic))
```
BovineOx, line 48: Is this adding in a bunch of reflection? This could become slow with large volumes

Converters/ArcGIS/Speckle.Converters.ArcGIS3/Utils/LocalToGlobalConverterUtils.cs
```diff
@@ -30,100 +31,82 @@ public Base TransformObjects(Base atomicObject, List<Matrix4x4> matrix)
...
-    // {
-    //
-    // }
-    // else if (newObject is SOG.Point point)
-    // {
-    //
-    // }
-    if (newObject is SOG.Mesh mesh)
+
+    if (atomicObject.TryGetDisplayValue() is IReadOnlyList<Base> listVals)
```
BovineOx, line 56: Interesting this Try method doesn't follow the standard C# pattern but instead returns this value if present?

Converters/ArcGIS/Speckle.Converters.ArcGIS3/Utils/LocalToGlobalConverterUtils.cs
```diff
@@ -30,100 +31,82 @@ public Base TransformObjects(Base atomicObject, List<Matrix4x4> matrix)
...
-    newObject["displayValue"] = newDisplayValue;
-    return newObject;
-  }
-
-  private SOG.Mesh TransformMesh(SOG.Mesh displayMesh, List<Matrix4x4> matrix)
-  {
-    List<List<double>> oldVertices = new();
-    for (int i = 0; i < displayMesh.vertices.Count; i += 3)
+    // copy the original object and assign new displayValue - hacky
+    Base newObject = Core.Api.Operations.Deserialize(Core.Api.Operations.Serialize(atomicObject));
```
BovineOx, line 101: This is creating a tight coupling to serialization we should avoid. Once, object versioning is in this is unlikely yo be viable - I would argue that it is beyond the scope of this method, probably this should become a service and we should avoid calling the static directly

BovineOx (changes requested):
I think there's some discussion, because it does appear to be a  bit of a hack and one that might break quite easily and become problematic with object versioning. It feels like there might be another option so might be worth discussing

Converters/ArcGIS/Speckle.Converters.ArcGIS3/Utils/LocalToGlobalConverterUtils.cs
```diff
@@ -28,102 +28,28 @@ public Base TransformObjects(Base atomicObject, List<Matrix4x4> matrix)
       return atomicObject;
     }
 
-    // This is a temp hack. We would add transformations to conversions later instead try to copy objects like this.
-    // Jedd also has opinions on this.
-    Base newObject = Core.Api.Operations.Deserialize(Core.Api.Operations.Serialize(atomicObject));
+    List<Objects.Other.Transform> transforms = new() { };
+    transforms.AddRange(matrix.Select(x => new Objects.Other.Transform(x, "none")).ToList());
```
oguzhankoral, line 32: Ideally same as you did but I would simplify it as one liner, and it is better. And also probably we do not necessarily have it as `ToList()` since we won't mutate it later, it is just a mapping one to another.
```c#
IEnumerable<Objects.Other.Transform> transforms = matrix.Select(x => new Objects.Other.Transform(x, "none"));
```

Converters/ArcGIS/Speckle.Converters.ArcGIS3/Utils/LocalToGlobalConverterUtils.cs
KatKatKateryna: possibly, but we want to preserve the original properties as much as possible

Build/packages.lock.json
```diff
@@ -38,9 +38,9 @@
       },
       "Speckle.InterfaceGenerator": {
         "type": "Direct",
-        "requested": "[0.9.6, )",
```
oguzhankoral, line 41: hmm, why we downgrade it? @adamhathcock is `0.9.6` version the fix you made for namespaces lately?

KatKatKateryna: locks are messed up, trying to resolve it

KatKatKateryna: I gave up, here is a PR build on the latest dev: https://github.com/specklesystems/speckle-sharp-connectors/pull/93

PR #88, opened 2024-07-30
feat(autocad): CNX-23 adds color and transparency as render materials on send and receive

clairekuang:
In autocad, we are sending every object's `color` and `transparency` as a render material (in lieu of not implementing display styles). This currently doesn't apply for layers, which already have a color.

This doesn't take into consideration, on receive, when the converted object is a list of display values, which may have their own render materials different from the parent which was passed to the converter. This will need to be debugged once revit send is polished.


clairekuang:
Dropping a note here: Curves in Rhino do **not** have render materials, and so receiving autocad curves with render materials (our fake display styles) will not display as expected.

PR #89, opened 2024-07-31
Logging using ILRepack and start of Open Telemetry usage

adamhathcock:


Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Receive/RevitHostObjectBuilder.cs
```diff
@@ -47,7 +48,12 @@ CancellationToken cancellationToken
   ) =>
     _syncToThread.RunOnThread(() =>
     {
-      var objectsToConvert = _traverseFunction.Traverse(rootObject).Where(obj => obj.Current is not Collection);
+      using var activity = SpeckleActivityFactory.Start("Build");
+      IEnumerable<TraversalContext> objectsToConvert;
+      using (var _ = SpeckleActivityFactory.Start("Traverse"))
```
JR-Morgan, line 53: Since traversal is just returning a lazily enumerated `IEnumerable` here, I can't see this activity being useful...

Sdk/Speckle.Connectors.Utils/Operations/ReceiveOperation.cs
```diff
@@ -29,26 +30,38 @@ IServerTransportFactory serverTransportFactory
...
+    HostObjectBuilderResult? res;
     // 2 - Check account exist
     Account account = _accountService.GetAccountWithServerUrlFallback(receiveInfo.AccountId, receiveInfo.ServerUrl);
-
-    // 3 - Get commit object from server
     using Client apiClient = new(account);
-    var version = await apiClient
-      .Version.Get(receiveInfo.SelectedVersionId, receiveInfo.ModelId, receiveInfo.ProjectId, cancellationToken)
-      .ConfigureAwait(false);
+    using (var _ = SpeckleActivityFactory.Start("Receive from server"))
```
JR-Morgan, line 40: a better message would be "Fetching Version Info"

Since this `Version` object is only lightweight info about the `Version`, not any object data.

JR-Morgan: I also wonder if we'd prefer this to be implemented inside the client....
as a "executing graphql request" trace on ALL graphql queries/mutations

Sdk/Speckle.Connectors.Utils/Operations/Send.cs
JR-Morgan, line 1: https://linear.app/speckle/issue/CXPLA-9/move-dui3-sendhelper-to-core-as-operationssend

Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Receive/RevitHostObjectBuilder.cs
```diff
@@ -56,31 +62,37 @@ CancellationToken cancellationToken
 
       var conversionResults = BakeObjects(objectsToConvert);
 
-      _transactionManager.CommitTransaction();
-      transactionGroup.Assimilate();
-
+      using (var _ = SpeckleActivityFactory.Start("Commit"))
```
JR-Morgan, line 65: "Commit transaction" would make more sense (to avoid confusion with the former naming of speckle `Version`s)

JR-Morgan (changes requested):

JR-Morgan (approved):
Approved, we'll polish the exact instrumentation usage later

JR-Morgan (approved):

PR #92, opened 2024-08-01
Ilrepack update with logging only

adamhathcock:


Sdk/Speckle.Connectors.Utils/Common/AssemblyExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+      var informationalVersion =
+        assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? "No version";
+      if (!string.IsNullOrEmpty(informationalVersion))
+      {
+        return informationalVersion;
+      }
+    }
+#pragma warning disable CA1031
+    catch (Exception)
```
JR-Morgan, line 19: Can we be more specific about the exceptions that we catch here?

Sdk/Speckle.Connectors.Utils/ContainerRegistration.cs
```diff
@@ -1,13 +1,77 @@
...
 using Speckle.Connectors.Utils.Cancellation;
 using Speckle.Connectors.Utils.Operations;
-using Speckle.Core.Logging;
+using Speckle.Logging;
+using ILogger = Microsoft.Extensions.Logging.ILogger;
 
 namespace Speckle.Connectors.Utils;
 
+#pragma warning disable CA1063
+public class SpeckleLoggerFactory : ILoggerFactory
```
JR-Morgan, line 12: Mark sealed

JR-Morgan (changes requested):

Sdk/Speckle.Connectors.Utils/Common/AssemblyExtensions.cs
JR-Morgan: Yes, but this is a copy paste from sentry, I'm happy for this to stay for now.

JR-Morgan (approved):

PR #93, opened 2024-08-01
AGIS: Blocks ToTransform for ITransformable

KatKatKateryna:


Converters/ArcGIS/Speckle.Converters.ArcGIS3/Utils/LocalToGlobalConverterUtils.cs
```diff
@@ -28,102 +29,43 @@ public Base TransformObjects(Base atomicObject, List<Matrix4x4> matrix)
...
+      foreach (var prop in atomicObject.GetMembers(DynamicBaseMemberType.Dynamic))
       {
-        group.Add(displayMesh.vertices[i + j]);
+        if (prop.Value is not Base)
+        {
+          atomicObject[prop.Key] = prop.Value;
+        }
+        else
+        {
+          // TODO: add more granular warnings if needed that the property was skipped for blocks
```
oguzhankoral, line 60: Why we prevent to add non base props? what if we have some attributes dictionary? I have it on sketchup, if you do it like this, you won't be able to get attributes that attached to blocks

KatKatKateryna: hmm, with Base I imagined a geometry being attached. Because those will not be transformed according to the matrices and might be confusing. 
Is there a specific "attributes" base object for blocks?

oguzhankoral: But still you might loose some info from original object. Why we need this if else statement at all?

KatKatKateryna: so when we have more granular reporting (e.g. Success status Warnings) we can notify the user

KatKatKateryna: we can do with the ticket! My impression is that allowing to pass Base "as is" to all instances will create more tech debt, as we cannot predict what the users will be passing. (e.g. someone creates a "TopRightCorner" property with a Point object, and it will be passed without transformations.)
https://linear.app/speckle/issue/CNX-153/agis-fill-the-placeholders-for-granular-reporting 
But ping me if you want to quickly align on this

oguzhankoral (dismissed):

oguzhankoral (approved):

PR #94, opened 2024-08-01
All operations: split main tread operations (ArcGIS, Rhino, Revit, Aacad) 

KatKatKateryna:
By now, Send and Receive operations (in particular RootObjectBuilder and HostObjectBuilder) were wrapped in SyncToThread function, which, in ArcGIS was redirecting the operation to the main thread using QueuedTask.Run from native host API. 

The problem with this approach was that, although Speckle UI is responsive, the main thread of the host app is blocked during the whole operation. And in ArcGIS, there are apparently some native host API subscriptions reacting on creation of new layers and features. Which means, after Receiving 10 layers in an occupied main thread, ArcGIS started running all of its own subscription events that took a long time, and made the impression that Speckle broke something in ArcGIS. 

With moving the operations away from SyncToThread, and making them async instead, we are able to send the individual conversions and baking to the main thread, returning the result and letting host API run it's stuff one by one. This delivers correct progress status to the user while still letting us run required functions on the main thread.


oguzhankoral:
@adamhathcock @KatKatKateryna could you please explain in PR description, what was the intention on removing `SyncToThread`. How it helps, what should we expect as behavior on host app? does it effect anything etc?

oguzhankoral (dismissed):
LGTM

oguzhankoral (approved):
Tested all connectors. Looks good. And confirming that there is no deadlock after receiving in arcgis.

PR #101, opened 2024-08-04
fix(dui3): cleans up base layer prefix from invalid chars - fixes second receive cleanup issues

didimitrie:
and adds extra defensibility around layer naming in acad. 

The issue was that we were not consistently cleaning up the base layer prefix: 
- layer creation would remove `/` chars
- base layer prefix would still contain `/` chars

resulting in all second receives from nested models being borked. 

Also fixes another acad second receive bug, where we would stop purging object layers if we would encounter one with no objects on it.


oguzhankoral (approved):
Tested 👌

PR #105, opened 2024-08-05
quick optimised rhino send

JR-Morgan:
I was doing some testing in Rhino to try and reproduce the long send problem,
I needed to send stuff quickly so I did some very low hanging fruit performance optimisations.

Should speed things up by a fair bit.


adamhathcock (approved):
Less allocations!

clairekuang (approved):
looks good!

PR #106, opened 2024-08-06
Feat(dui3) Debouncing progress updates

oguzhankoral:
My initial attempt was on Timers, but I had random behaviors on Autocad, then decided to go with percentage difference checks. Since we round decimals on UI, we don't need to send progress updates less than %1. OperationProgressManager handles it with concurrent dictionary. Open to suggestions, I might do more tests before merging it.

### DIFF
--- a/DUI3/Speckle.Connectors.DUI/Bindings/OperationProgressManager.cs
+++ b/DUI3/Speckle.Connectors.DUI/Bindings/OperationProgressManager.cs
@@ -1,20 +1,17 @@
-﻿// using System.Collections.Concurrent;
-// using System.Diagnostics;
+﻿using System.Collections.Concurrent;
 using Speckle.Connectors.DUI.Bridge;
 using Speckle.Connectors.DUI.Models.Card;
 
 namespace Speckle.Connectors.DUI.Bindings;
 
 /// <summary>
-/// Debouncing to progress update for UI.
+/// Debouncing progress for every %1 update for UI.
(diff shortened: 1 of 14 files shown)


DUI3/Speckle.Connectors.DUI/Bindings/OperationProgressManager.cs
```diff
@@ -0,0 +1,48 @@
...
+using Speckle.Connectors.DUI.Bridge;
+using Speckle.Connectors.DUI.Models.Card;
+
+namespace Speckle.Connectors.DUI.Bindings;
+
+/// <summary>
+/// Debouncing progress for every %1 update for UI.
+/// This class requires a specific bridge in its binding, so registering it will create random bridge which we don't want to.
+/// </summary>
+public class OperationProgressManager
```
adamhathcock, line 11: I would make this class a singleton and inject it on usage.  The Bridge isn't needed until set is called so doesn't need to be on the constructor.

How does it know it's done?   Does the diff only care if it's changed, not if it's greater?

adamhathcock (changes requested):

DUI3/Speckle.Connectors.DUI/Bindings/OperationProgressManager.cs
oguzhankoral: that's a good point

adamhathcock (approved):
Smells better

PR #114, opened 2024-08-07
Made Conversion Results no longer inherit `Base`

JR-Morgan:
Due to the Sdk changes & re-namespacing (see https://www.notion.so/speckle/Re-namespacing-of-3-x-SDK-for-2-x-side-by-side-b70e768915014fc6a19f44db7e46c8ec)
all base subclasses now need an explicit speckle_type attribute.

I didn't want to have to spark up the discussion about "what should the speckle type of conversion results be" since I don't think we have a clear picture of how this feature should work.

After agreeing with @didimitrie, for the September launch we'll not send conversion results in the current way.
We can chat later to decide what we should do, and if needed we can make them `Base` subclasses again.

### DIFF
--- a/Sdk/Speckle.Connectors.Utils/Connector.cs
+++ b/Sdk/Speckle.Connectors.Utils/Connector.cs
@@ -1,4 +1,4 @@
-﻿using Speckle.Objects.Structural.GSA.Geometry;
+using Speckle.Objects.Geometry;
 using Speckle.Sdk;
 using Speckle.Sdk.Host;
 using Speckle.Sdk.Logging;
@@ -10,7 +10,7 @@ public static class Connector
 {
   public static IDisposable? Initialize(HostApplication application, HostAppVersion version)
   {
-    TypeLoader.Initialize(typeof(Base).Assembly, typeof(GSAAssembly).Assembly);
+    TypeLoader.Initialize(typeof(Base).Assembly, typeof(Point).Assembly);
     var config = new SpeckleConfiguration(
       application,
       version,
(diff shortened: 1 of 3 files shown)


JR-Morgan:
CI will fail on this until https://github.com/specklesystems/speckle-sharp-sdk/pull/59 is merged.
This is fine, because we are targeting #113

Sdk/Speckle.Connectors.Utils/Connector.cs
```diff
@@ -10,7 +10,7 @@ public static class Connector
 {
   public static IDisposable? Initialize(HostApplication application, HostAppVersion version)
   {
-    TypeLoader.Initialize(typeof(Base).Assembly, typeof(GSAAssembly).Assembly);
+    TypeLoader.Initialize(typeof(Base).Assembly, typeof(Point).Assembly);
```
didimitrie, line 13: looks tangential to this pr right?

didimitrie (approved):
see my comment above though!

Sdk/Speckle.Connectors.Utils/Connector.cs
JR-Morgan: Yes, sorry it became somewhat irrelevant, this is how we're registering which assemblies contain Base subtypes.

I would have needed to add Speckle.Connectors.Utils here (via a `typeof(Conversion Result).Assembly`

My change from GSAAssembly to Point was simply to make it easier to see that we're getting the Objects assembly

PR #117, opened 2024-08-07
feat(autocad): CNX-23 render materials on send and receive

clairekuang:
Adds render materials to autocad on send and receive.
This is a new functionality for Autocad dui3 (v2 did not have materials support)!
Also includes smol refactor of `AutocadGroupUnpacker` to reduce complexity

Some material conversions don't work well (top row: original, bottom row: received), most notably the global default material which gives a black diffuse color when extracted. This could probably be fixed later. Like Rhino, to get better material interop, we should be storing the native material type and other fields as well.

image

#88 can be closed without merging, once this is merged in


Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadLayerManager.cs
```diff
@@ -67,14 +68,30 @@ public Layer GetOrCreateSpeckleLayer(Entity entity, Transaction tr, out LayerTab
...
     {
-      string layerId = layerPath[j].applicationId ?? layerPath[j].id;
-      if (objectColorIdMap.TryGetValue(layerId, out layerColor))
+      bool foundColor = false;
+      bool foundMaterial = false;
+      for (int j = layerPath.Length - 1; j >= 0; j--)
       {
-        break;
+        string layerId = layerPath[j].applicationId ?? layerPath[j].id;
+        if (foundColor && foundMaterial)
```
oguzhankoral, line 81: why we break at the next iteration? we could make it more readable as below

```c#
    // Try to find color
    if (!foundColor && objectColorsIdMap.TryGetValue(layerId, out layerColor))
    {
      foundColor = true;
    }
    // Try to find material
    if (!foundMaterial && objectMaterialsIdMap.TryGetValue(layerId, out layerMaterial))
    {
      foundMaterial = true;
    }

    // Break early if both are found
    if (foundColor && foundMaterial)
    {
      break;
    }
```

oguzhankoral (changes requested):
Minor

oguzhankoral (approved):
Code-wise LGTM

oguzhankoral (approved):

Connectors/Rhino/Speckle.Connectors.RhinoShared/Operations/Receive/RhinoHostObjectBuilder.cs
```diff
@@ -208,8 +208,6 @@ CancellationToken cancellationToken
             conversionResults.Add(new(Status.SUCCESS, obj, r, result.GetType().ToString()));
             bakedObjectIds.Add(r);
           }
-
-          applicationIdMap[objectId] = conversionIds;
```
didimitrie, line 212: @clairekuang this was bad juju change - wondering why it happened? I'll fix it on my end now (in this branch)

oguzhankoral (approved):

didimitrie (approved):

PR #119, opened 2024-08-07
CNX-193 Track layer changes by ModifyObjectAttributes of RhinoObject

oguzhankoral:
- `RhinoDoc` has an event to track changes on `ObjectAttributes` on `RhinoObject` level, only `LayerIndex` considered as change, we can enrich it later according to needs.
- AutoCAD already tracks layer changes.

Rhino_qGoLoBheFX

acad_IEVrlwhGoi


didimitrie (approved):
another nice one 💖 - this will be as well good for when we will want to detect changes to other attributes too, tbh 🤘

PR #123, opened 2024-08-08
Dim/cnx 23 fixes

didimitrie:
Cleanp & simplifications. tl;dr: 

**Rhino**: 
- **FIX: instance creation was broken because of unwanted non-population of application id map**
- simplifies material assigning logic by creating a flat object <> material proxy map on material conversion, and removes a lotta indirect redundant code as a result; 
- layer manager is now responsible for assigning layer color/mat on creation; 
- object color/material assigning is simpler now from host object builder

**Acad:** 
- nothing really to change in here, stuff was as simple as it could get already imho

### DIFF
--- a/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadLayerManager.cs
+++ b/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadLayerManager.cs
@@ -13,6 +13,8 @@ namespace Speckle.Connectors.Autocad.HostApp;
 public class AutocadLayerManager
 {
   private readonly AutocadContext _autocadContext;
+  private readonly AutocadMaterialManager _materialManager;
+  private readonly AutocadColorManager _colorManager;
   private readonly string _layerFilterName = "Speckle";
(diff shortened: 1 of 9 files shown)


Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadLayerManager.cs
```diff
@@ -56,9 +64,9 @@ public Layer GetOrCreateSpeckleLayer(Entity entity, Transaction tr, out LayerTab
   /// <returns>The name of the existing or created layer</returns>
   public string CreateLayerForReceive(
     Collection[] layerPath,
-    string baseLayerPrefix,
-    Dictionary<string, AutocadColor> objectColorsIdMap,
-    Dictionary<string, ObjectId> objectMaterialsIdMap
+    string baseLayerPrefix
+  // Dictionary<string, AutocadColor> objectColorsIdMap,
```
oguzhankoral, line 68: please remove it not used anymore

oguzhankoral (approved):

PR #126, opened 2024-08-08
Dimitrie/cnx 201 group names from autocad to rhino are not supported

didimitrie:
Fixes group creation so they are correctly named; this required purging previously created groups for second receive workflows. I've taken the chance to also move a lot of the group creation logic outside of the host object builder to the group manager itself, though IIRC the group manager was meant to be split into group unpacker and group baker, but it got merged into one in the last PR on render materials. 

See [CNX-201](https://linear.app/speckle/issue/CNX-201/group-names-from-autocad-to-rhino-are-not-supported) for full details of the issue.

Proof it works for Bilal:
image.


oguzhankoral (approved):
Tested, works nice!

PR #127, opened 2024-08-09
fix(rhino): CNX-245 render materials inherited from layer bug

clairekuang:
Fixes bug where the first created render material was being used instead of none assigned when no render material is found on an object.

image


oguzhankoral (approved):
LGTM

PR #132, opened 2024-08-12
Adam/cnx 249 add autocad 2025 connector and converter

adamhathcock:
https://linear.app/speckle/issue/CNX-249/add-autocad-2025-connector-and-converter


clairekuang (approved):
lgtm

PR #136, opened 2024-08-12
CXPLA-17 Add XX_OR_GREATER flags to converter/connector projects

AlanRynne:
Adds XX_OR_GREATER to all converter and connector projects:

- Rhino
- Revit
- Acad/Civil

Left ArcGIS out because it has no other version, so it's intention is not as clear.

### DIFF
--- a/Converters/Civil3d/Speckle.Converters.Civil3d2024/Speckle.Converters.Civil3d2024.csproj
+++ b/Converters/Civil3d/Speckle.Converters.Civil3d2024/Speckle.Converters.Civil3d2024.csproj
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TargetFramework>net48</TargetFramework>
     <PlatformTarget>x64</PlatformTarget>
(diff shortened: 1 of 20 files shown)


Converters/Civil3d/Speckle.Converters.Civil3d2024/Speckle.Converters.Civil3d2024.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TargetFramework>net48</TargetFramework>
     <PlatformTarget>x64</PlatformTarget>
+    <DefineConstants>$(DefineConstants);CIVIL3D;CIVIL3D2024;CIVIL3D2024_OR_GREATER</DefineConstants>
```
JR-Morgan, line 6: Guessing we should have `CIVIL3D2023_OR_GREATER` here too?
Tho, not sure the status of our CIVIL versions... if we don't have other versions yet, then we should just remember to correct the `DefineConstants` later

JR-Morgan (approved):

PR #142, opened 2024-08-13
feat(revit): extracts collection creation logic out into SendCollectionManager

didimitrie:
Several things: 
- Cleans up root object builder a bit in revit. Had potential to get messy. 
- Fully cleans up hosted element convertor


oguzhankoral (approved):
Nice!

PR #146, opened 2024-08-13
fix(autocad): CNX-274 circulararc conversion now handles edge case

clairekuang:
Previously some polylines with arc segments were resulting in the arc segments being converted with a midpoint = endpoint. This was resulting in failed conversions when receiving in rhino.

I've added some comments on why this seems to happen, as well as an extra check in the arc conversion to use a re-parametrized domain for midpoint calculation.

I've also added a new exception thrown in the rhino polycurve conversion when any segment is not converted to a valid curve.

Tested with multiple autocad files, and receiving them in rhino.


oguzhankoral (approved):
Tested through acad, rhino, sketchup

PR #147, opened 2024-08-14
Agis colors receive

KatKatKateryna:
Receive ColorProxies in ArcGIS:
- create UniqueRenderer for every Feature Layer
- create unique color groups for each object in the layer, associate color with Speckle_ID
- if colorProxy not found for the object, check if any of the object parents has the color assigned and apply it.


KatKatKateryna:
needs more commits to test

Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/HostApp/ArcGISColorManager.cs
```diff
@@ -43,6 +48,151 @@ public List<ColorProxy> UnpackColors(List<(MapMember, int)> mapMembersWithDispla
...
+
+    // declare default white color
+    Color color = Color.FromArgb(255, 255, 255, 255);
+
+    // get color moving upwards from the object
+    foreach (var parent in tc.GetAscendants())
+    {
+      if (parent.applicationId is string appId && ObjectColorsIdMap.TryGetValue(appId, out Color objColor))
+      {
+        color = objColor;
```
KatKatKateryna, line 89: Should we get the color of the closest parent, or the highest in the hierarchy?

Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/HostApp/ArcGISColorManager.cs
```diff
@@ -43,6 +48,151 @@ public List<ColorProxy> UnpackColors(List<(MapMember, int)> mapMembersWithDispla
...
+        // POC: should we get the color of the closest parent and break, or walk up to the highest possible?
+        break;
+      }
```
KatKatKateryna, line 92: Should we get the color of the closest parent, or the highest in the hierarchy?

clairekuang: Not sure I understand the question here: if the atomic object doesn't have a color, then the color of the closest parent with a color should be used

clairekuang (dismissed):
did some testing receiving from rhino, looks good!
However, receiving render material proxies should exist as well, for receiving from applications where no colors are being sent (eg revit).
This logic could just check for ["colorProxies"] on the base collection, if none exist, check for ["renderMaterialProxies"] instead and use those values.

Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/HostApp/ArcGISColorManager.cs
clairekuang: Let's resolve this POC comment before merge

clairekuang (approved):

PR #149, opened 2024-08-14
Dim/revit materials

didimitrie:
[CNX-186: Proper material support for receiving in Rhino, Sketchup and co.](https://linear.app/speckle/issue/CNX-186/proper-material-support-for-receiving-in-rhino-sketchup-and-co)

the tl;dr: 

Revit generates from one object multiple display values, all with their own render materials. We now correctly send render material proxies out of revit with this information, and correctly view/assign them in rhino, sketchp and acad (arcgis is a todo that i will need to delegate). 

**To do so, there are several changes:** 

On send, in revit: 
- each display value mesh from revit gets its own auto generated application id.
- the render material proxies are generated by the revit's `MeshByMaterialDictionaryToSpeckle` converter
- the revit converter context stack now keeps a singleton cache of these render material proxies that persists across send operations. note it needs to have this scope as caching at the connector level bypasses the converter. 

On receive: 
- the fallback converter's generic `object` return can now mean two things:
  - 1-1 scenario: a single object
  - 1-many scenario: an enumerable of (originalDisplayValueBase, hostAppDisplayValueObject)
- this is required to enable correct material mapping between display values with multiple materials
- I've converged both rhino and acad on a more opionionated and clear path of `BakeObject` and `BakeObjectAsGroups`.

Related changes: 
- The display value extractor class now stops caring about transforms
- Removed an unused dependency (conversion cache ported from 2.0) 

This PR also includes some other unrelated fly by fixes: 
- Acad: correctly escape material names
- Acad: correctly escape group names
- Revit: progress reporting going > 100%
- Revit: [CNX-125: Autodesk.Revit.Exceptions.InvalidOperationException: Can not subscribe to an event during execution of that event!](https://linear.app/speckle/issue/CNX-125/autodeskrevitexceptionsinvalidoperationexception-can-not-subscribe-to)

**NOTE: Do not merge without patching ArcGIS, as it will break due to the fallback converter's new behaviour.**


Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/Raw/Geometry/MeshByMaterialDictionaryToSpeckle.cs
```diff
@@ -5,7 +5,7 @@
 namespace Speckle.Converters.RevitShared.ToSpeckle;
 
 public class MeshByMaterialDictionaryToSpeckle
-  : ITypedConverter<Dictionary<DB.ElementId, List<DB.Mesh>>, List<SOG.Mesh>>
+  : ITypedConverter<(Dictionary<DB.ElementId, List<DB.Mesh>> target, DB.ElementId parentElementId), List<SOG.Mesh>>
```
oguzhankoral, line 8: these types are getting too much.. They started to deserve a wrapper class for a purpose

Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/Raw/Geometry/MeshByMaterialDictionaryToSpeckle.cs
```diff
@@ -34,15 +34,25 @@ public MeshByMaterialDictionaryToSpeckle(
...
-  public List<SOG.Mesh> Convert(Dictionary<DB.ElementId, List<DB.Mesh>> target)
+  public List<SOG.Mesh> Convert((Dictionary<DB.ElementId, List<DB.Mesh>> target, DB.ElementId parentElementId) args)
   {
-    var result = new List<SOG.Mesh>(target.Keys.Count);
+    var result = new List<SOG.Mesh>(args.target.Keys.Count);
+    var objectRenderMaterialProxiesMap = _contextStack.RenderMaterialProxyCache.ObjectRenderMaterialProxiesMap;
 
-    foreach (var meshData in target)
+    var materialProxyMap = new Dictionary<string, RenderMaterialProxy>();
+    objectRenderMaterialProxiesMap[args.parentElementId.ToString()!] = materialProxyMap;
```
oguzhankoral, line 45: how confident we are, we always reset the list of material proxies of `parent` in the cache?

oguzhankoral (changes requested):

oguzhankoral (approved):

PR #156, opened 2024-08-15
Cleaned up cancellation token source usages

JR-Morgan:
We shouldn't really be passing around `CancellationTokenSources`, since they are designed (as all IDisposables are designed) to be managed by a single owning object.
Instead passing cancellationTokens to observe a cancel is preffered.


oguzhankoral (approved):

AlanRynne (approved):
Fantastic ♥️

PR #158, opened 2024-08-15
fix(rhino): group selection on highlight + get send settings

didimitrie:
Two fast fixes: 

(1) Group selection throwing: 
Models received in rhino from revit would throw, as `object.GetGroupList()` actually returns null if empty:

```
Allocates an array of group indices of length GroupCount. If GroupCount is 0, then this method returns null.
```

(2) GetSendSettings in rhino was a property, made it a function and disabled the compiler error. The bridge needs methods.


Connectors/Rhino/Speckle.Connectors.RhinoShared/Bindings/RhinoSendBinding.cs
```diff
@@ -145,7 +145,12 @@ private void SubscribeToRhinoEvents()
 
   public List<ISendFilter> GetSendFilters() => _sendFilters;
 
-  public List<CardSetting> GetSendSettings() => [];
+#pragma warning disable CA1024
```
oguzhankoral, line 148: @JR-Morgan do we still need this pragma and `return new()`? Isn't it `[]` is same?

JR-Morgan: [CA1024: Use properties where appropriate](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1024) will not complain so long as your implementing an interface.

So it can be removed here, unrelated to the list stuff.

But fyi, `[]` will be have the same as` new List<T>()` for list typed variables/returns.

PR #167, opened 2024-08-19
fix(autocad): CNX-318 better color inheritance for blocks

clairekuang:
Because block definition object colors can inherit their colors by layer or by block, and block instance colors can also be by layer or by block, this pr is setting the color for block definition objects as if they are by color in both conditions, to simulate the correct colors in the viewer (and hopefully when receiving in other apps).
NOTE: also sending a `source` string prop on the color proxies to identify propery color inheritance on receive

includes a refactor for speckle application ids by adding extension methods for layers, entities, materials, colors, etc to guarantee we are always using the same app id.

Original autocad file: 
image

Received (old): https://app.speckle.systems/projects/b53a53697a/models/66aca735ab@0836579d97
image

Received (new): 
https://app.speckle.systems/projects/b53a53697a/models/66aca735ab@55ecf653c7
image


didimitrie:
> includes a refactor for speckle application ids by adding extension methods for layers, entities, materials, colors, etc to guarantee we are always using the same app id.

finally! this is super nice to have!

oguzhankoral:
I suggest to use another proxy class for `ColorSource`. By this way, viewer also can know which source it will apply to each object, layer? block? etc.. And I believe this is the clearest approach. Otherwise we're not telling anything to viewer render any geometry from what source. WDYT @didimitrie ?

```json
{
  "colorSourceProxies": [
    {
      "speckle_type": "ColorSourceProxy",
      "name": "ByLayer",
      "objects": ["12345", "12346", "12347", "12348"]
    },
    {
      "speckle_type": "ColorSourceProxy",
      "name": "ByBlock",
      "objects": ["22345", "22346", "22347", "22348"]
    }
  ]
}
```

clairekuang:
> I suggest to use another proxy class for `ColorSource`. By this way, viewer also can know which source it will apply to each object, layer? block? etc.. And I believe this is the clearest approach. Otherwise we're not telling anything to viewer render any geometry from what source.

Currently I'm attaching the source dynamically to the colorproxy: 
image
I think this is fine for now, source can have three values (`block`, `layer`, `object`) and is factored into the color proxy id, and this should also render properly in the current viewer logic

didimitrie (approved):

oguzhankoral (approved):

PR #170, opened 2024-08-20
Agis separate dataset creation

KatKatKateryna:
Refactoring HostObjectBuilder to treat all features individually, including GisFeatures. 
This allows to treat them in the same way in the steps of grouping into layers and Writing to datasets.
This also resolves the issue of the Map being stuck with the outdated dataset representation even after receiving the new version and overwriting it (because now we return back to main thread after clearing old datasets)


Converters/ArcGIS/Speckle.Converters.ArcGIS3/ToHost/TopLevel/GisMultipatchFeatureToHostConverter.cs
```diff
@@ -0,0 +1,56 @@
...
+  {
+    _polygon3dConverter = polygon3dConverter;
+    _multipatchConverter = multipatchConverter;
+  }
+
+  public object Convert(Base target) => Convert((SGIS.GisMultipatchFeature)target);
+
+  public ACG.Geometry Convert(SGIS.GisMultipatchFeature target)
+  {
+    List<SGIS.PolygonGeometry3d> polygonGeometry = target
```
adamhathcock, line 34: Not sure this is the most optimial usage.  ToList is expensive.

Probably can just loop once over the target.geometry and switch on pattern matching casts to do a count.

Converters/ArcGIS/Speckle.Converters.ArcGIS3/Utils/CharacterCleaner.cs
```diff
@@ -1,7 +1,9 @@
 using System.Text.RegularExpressions;
+using Speckle.InterfaceGenerator;
 
 namespace Speckle.Converters.ArcGIS3.Utils;
 
+[GenerateAutoInterface]
```
adamhathcock, line 6: Nice!

adamhathcock (changes requested):
I didn't read everything in detail but there should probably be some optimization when using collections and Linq to avoid slowness.

I'm enjoying the usage of Interfaces and DI!

adamhathcock (dismissed):

AlanRynne (approved):
Kat ran me through the functionality changes on friday, and I see no glaring issues code-wise either 🙂

PR #171, opened 2024-08-20
fix(rhino): CNX-319 rhino block colors

clairekuang:
Fixes rhino color inheritence on send and receive by adding source information to color proxies.
Also fixes bug on instance bake where we were forgetting to set colors and materials (since instance bake happens in the instance manager instead of the host obj builder)

Rhino test file:
image

Before: https://app.speckle.systems/projects/93200a735d/models/cbacd3eaeb@344a397239
image

After: https://app.speckle.systems/projects/93200a735d/models/d1fbf678f1@50491295f9
image


didimitrie (dismissed):

oguzhankoral (dismissed):

oguzhankoral (approved):

clairekuang:
Waiting for coordination with @AlexandruPopovici before merging, to see if definition object layer hack needs to exist for rhino.

PR #172, opened 2024-08-20
AGIS receive obj without appId

KatKatKateryna:


Sdk/Speckle.Connectors.Utils/Instances/LocalToGlobalUnpacker.cs
```diff
@@ -45,9 +45,9 @@ List<TraversalContext> objectsToUnpack
     {
       if (atomicObject.applicationId is null)
       {
-        continue;
+        objectsAtAbsolute.Add((tc, atomicObject)); // to bake objs without appId, e.g. GH
```
didimitrie, line 48: L48 and L59 are the same, which makes me think this if can be restructured somehow, but no brainpower right now to say how.

didimitrie (changes requested):

Sdk/Speckle.Connectors.Utils/Instances/LocalToGlobalUnpacker.cs
didimitrie: okay had some brainpower post coffee: https://github.com/specklesystems/speckle-sharp-connectors/pull/175

oguzhankoral (approved):

PR #173, opened 2024-08-20
feat(rhino): pre bakes layers on receive for speed and profit

didimitrie:
[CNX-325: Rhino: Change receive operation order to increase performance](https://linear.app/speckle/issue/CNX-325/rhino-change-receive-operation-order-to-increase-performance)

To test: receive this https://latest.speckle.systems/projects/126cd4b7bb/models/afec5fcb04 before and after, with the rhino tab panels open.


clairekuang (approved):
nice

PR #180, opened 2024-08-20
Uses tag but needs version specific stuff

adamhathcock:
uses minver again


AlanRynne:
Rejecting since we figured out how to move forward with GitVersion

PR #181, opened 2024-08-20
fix(revit): CNX-329 adds units to all relevant conversions

clairekuang:
All non-geometry conversions were missing units: this pr adds them in


clairekuang:
Display meshes should already have units on them  if the mesh converter is used, if not then definitely yes! I haven’t added them to RevitElements because the class doesn’t have a units property, but can add them dynamically 🤔

didimitrie (approved):

PR #184, opened 2024-08-21
feat(revit): CNX-188 extract material quantities out for all objects

clairekuang:
Adds material quantities to all revit objects. Smaple commit: https://app.speckle.systems/projects/93200a735d/models/1b29238359@ae1b38c9aa

This is done by mirroring behavior in V2, by:
- converting material quantities per element
- converting and caching material per material quantity

Also fixes a bug from display value extractor for lighting fixtures that was preventing commits from being sent.

However, for post-september release we should seriously consider sending `MaterialProxies` instead of attached the material to every material quantity. Need to validate how this affect data workflows in powerbi, viewer, and excel.


Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/Raw/MaterialQuantitiesToSpeckle.cs
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    var (solids, _) = _displayValueExtractor.GetSolidsAndMeshesFromElement(element, null);
+
+    (double area, double volume) = GetAreaAndVolumeFromSolids(solids);
+    return CreateMaterialQuantity(element, material.Id, area, volume);
+  }
+
+  //Retrieves the revit material from assigned system type for mep elements
+  private static DB.Material? GetMEPSystemRevitMaterial(DB.Element e)
```
oguzhankoral, line 91: private static?

didimitrie (approved):

Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/Raw/MaterialQuantitiesToSpeckle.cs
```diff
@@ -0,0 +1,130 @@
...
+
+/// <summary>
+/// NOTE: Phased out (for now) due to dependency on the MaterialQuantity class, which creates/promotes a bloated data extraction.
+/// </summary>
+[Obsolete("Creates a rather bloated data structure - 2.0 style. More in the comment above.")]
+public class MaterialQuantitiesToSpeckle : ITypedConverter<DB.Element, List<MaterialQuantity>>
+{
+  private readonly ITypedConverter<DB.Material, (RevitMaterial, RenderMaterial)> _materialConverter;
+  private readonly RevitMaterialCacheSingleton _materialCacheSingleton;
+  private readonly DisplayValueExtractor _displayValueExtractor;
```
oguzhankoral, line 70: seems like these dudes shouldn't be injected at all

Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/Raw/MaterialQuantitiesToSpeckle.cs
```diff
@@ -0,0 +1,130 @@
...
+  {
+    _scalingService = scalingService;
+    _contextStack = contextStack;
+  }
+
+  /// <summary>
+  ///
+  /// </summary>
+  /// <param name="target"></param>
+  /// <returns></returns>
```
oguzhankoral, line 27: nice doc
jokes aside, it might be good to explain why it is list of dict as key string as value object?

oguzhankoral (changes requested):
minors

oguzhankoral (approved):

PR #189, opened 2024-08-22
fix(revit): invalidates send cache on doc unit change

didimitrie:
[CNX-350: Invalide send hash on Document Units Change](https://linear.app/speckle/issue/CNX-350/invalide-send-hash-on-document-units-change)

### DIFF
--- a/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
+++ b/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
@@ -188,8 +188,46 @@ internal sealed class RevitSendBinding : RevitBaseBinding, ISendBinding
       ChangedObjectIds[elementId.ToString()] = 1;
     }
 
-    // TODO: CHECK IF ANY OF THE ABOVE ELEMENTS NEED TO TRIGGER A FILTER REFRESH
+    if (HaveUnitsChanged(e.GetDocument()))
+    {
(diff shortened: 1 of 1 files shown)


Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
```diff
@@ -188,10 +188,48 @@ private void DocChangeHandler(Autodesk.Revit.DB.Events.DocumentChangedEventArgs
...
+    if (HaveUnitsChanged(e.GetDocument()))
+    {
+      var objectIds = Store.GetSenders().SelectMany(s => s.SendFilter != null ? s.SendFilter.GetObjectIds() : []);
+      _sendConversionCache.EvictObjects(objectIds);
+    }
     _idleManager.SubscribeToIdle(nameof(RevitSendBinding), RunExpirationChecks);
   }
 
+  // Keeps track of doc and current units
+  private readonly Dictionary<string, string> _docUnitCache = new();
```
oguzhankoral, line 200: if you switch multiple times in same documents it can cause the return false from `HaveUnitsChanged` by any chance? I don't see anywhere we remove item from this dict

oguzhankoral (changes requested):

Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
oguzhankoral: oh wait, you override the key

oguzhankoral (approved):

PR #193, opened 2024-08-22
supports points as fallback in Rhino

KatKatKateryna:
Resolving Receive of ArcGIS Point features
https://linear.app/speckle/issue/CNX-360/arcgis2rhino-points-not-received


oguzhankoral (approved):
LGTM

PR #194, opened 2024-08-22
chore(autocad/rhino): CNX-357 removes unnecessary definition color logic

clairekuang:
After a sync with @AlexandruPopovici , we've decided that the viewer will implement color logic to take into account display behavior by source. 

This behavior is outlined in detail in: https://www.notion.so/speckle/Objects-Colors-3a16d721e9fd46bda5e1b31072a38f29?pvs=4

PR removes hacky logic in autocad and rhino to send definition object colors `by layer` as if `by object`. There is no change in behavior in autocad<>autocad, rhino<>rhino, and autocad<>rhino workflows for color.


oguzhankoral (approved):
Nice!

PR #197, opened 2024-08-24
Dimitrie/cnx 358 caching element from another model sent

didimitrie:
There were several compounding issues that this PR is resolving: 
- harmonises applicationId mismatch between caching logic and converter by centralising on UniqueIds
- correctly invalidates cache in expiry checks by transforming elementIds into uniqueIds

### DIFF
--- a/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
+++ b/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
@@ -234,10 +234,22 @@ internal sealed class RevitSendBinding : RevitBaseBinding, ISendBinding
   {
     var senders = Store.GetSenders();
     string[] objectIdsList = ChangedObjectIds.Keys.ToArray();
+    var doc = RevitContext.UIApplication?.ActiveUIDocument.Document;
+
+    if (doc == null)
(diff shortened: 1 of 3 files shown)


Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
```diff
@@ -244,10 +244,23 @@ private void RunExpirationChecks()
   {
     var senders = Store.GetSenders();
     string[] objectIdsList = ChangedObjectIds.Keys.ToArray();
```
oguzhankoral, line 246: Why this can't return UniqueId already?

oguzhankoral: I mean why we don't  store unique ids already

oguzhankoral: OK. I got it. API returns element ids from events

didimitrie: :D Yes

oguzhankoral (approved):

PR #199, opened 2024-08-24
Fix(rhino): Do not skip first item if it is one-to-one conversion

oguzhankoral:
It was on me. While fixing the group highlight performance issue, this scenario skipped..


didimitrie (approved):

didimitrie:
this was my bad for post multi material, thx for fixing!

PR #202, opened 2024-08-26
Fix(revit): regression on switching ElementId to UniqueId

oguzhankoral:
It was a regression on switch from ElementId to UniqueId, since we send report items as UniqueId, it can't parse them directly to ElementId

### DIFF
--- a/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
+++ b/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
@@ -130,10 +130,14 @@ internal sealed class RevitSendBinding : RevitBaseBinding, ISendBinding
         }
       );
 
+      var activeUIDoc =
+        RevitContext.UIApplication?.ActiveUIDocument
+        ?? throw new SpeckleException("Unable to retrieve active UI document");
+
       List<ElementId> revitObjects = modelCard
         .SendFilter.NotNull()
         .GetObjectIds()
-        .Select(ElementIdHelper.Parse)
+        .Select(uid => ElementIdHelper.GetElementIdByUniqueId(activeUIDoc.Document, uid))
         .ToList();
 
       if (revitObjects.Count == 0)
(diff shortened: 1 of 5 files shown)


Connectors/Revit/Speckle.Connectors.RevitShared/ElementIdHelper.cs
```diff
@@ -5,13 +5,14 @@ namespace Speckle.Connectors.RevitShared;
 
 public static class ElementIdHelper
 {
-  public static ElementId Parse(string idStr)
+  public static ElementId GetElementIdByUniqueId(Document doc, string uniqueId)
```
didimitrie, line 8: I'd name this `GetElementIdFromUniqueId`, as we're transmuting an unique id into an element id, and in my head sounds clearer

Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
```diff
@@ -130,10 +130,14 @@ public async Task Send(string modelCardId)
...
 
+      var activeUIDoc =
+        RevitContext.UIApplication?.ActiveUIDocument
+        ?? throw new SpeckleException("Unable to retrieve active UI document");
+
       List<ElementId> revitObjects = modelCard
         .SendFilter.NotNull()
         .GetObjectIds()
-        .Select(ElementIdHelper.Parse)
+        .Select(uid => ElementIdHelper.GetElementIdByUniqueId(activeUIDoc.Document, uid))
```
didimitrie, line 140: i think there's some unecessary bouncing around here, from unique ids to element ids: in the root object builder, we can just use [GetElement(string)](https://www.revitapidocs.com/2019/b5a1473c-21c4-6c29-f1cf-26822c955260.htm) - where the string is the unique id.

didimitrie (changes requested):
briefly tested, and it works as advertised; minor request only. I might do PR

Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
oguzhankoral: The helper function already do it but yes, simplified as you suggested

didimitrie (approved):

PR #203, opened 2024-08-27
feat(revit): CNX-340 add reference point setting to revit send

clairekuang:
Adds reference point setting to revit send to mirror V2 behavior.
This still bakes in the selected reference point transform into commit geometry, and does not attach settings to the main commit.

Notes:
- Removes previously unused RevitConverterSettings class from converter.
- **!! Removes xyz converter test due to complications mocking the ReferencePointConverter !!**

sample commit comparing the same wall sent via project base point vs survey point (has translation and rotation):
https://app.speckle.systems/projects/b53a53697a/models/d7f7cf1df9#diff={%22diffs%22:[{%22versionA%22:%22d7f7cf1df9@1c1e492496%22,%22versionB%22:%22d7f7cf1df9@c7725813ef%22}]}


Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
```diff
@@ -103,11 +114,92 @@ fidelityString is not null
...
       }
       _detailLevelCache[modelCard.ModelCardId.NotNull()] = fidelity;
-      return new ToSpeckleSettings(fidelity);
+      return fidelity;
     }
+
     throw new ArgumentException($"Invalid geometry fidelity value: {fidelityString}");
   }
 
+  private Transform? GetReferencePointSetting(SenderModelCard modelCard)
```
oguzhankoral, line 123: A couple of question/comment around this and change requests accordingly;

1. Do these 3 types of reference points always generate same transforms in model same or not? If so why we calculate them always? I would cache them.

2. `SendBinding` started to bubble up, nice to have if we extract "GetSendSettings" to outside

oguzhankoral (changes requested):

Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
clairekuang: > Do these 3 types of reference points always generate same transforms in model same or not? If so why we calculate them always? I would cache them.

Unfortunately we aren't tracking if the reference points in the doc are changed between send operations, so for now the transform needs to be recalculated every time. This should be *ok* performance-wise bc it should only happen once per send operation.

Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/Settings/ReferencePointSetting.cs
```diff
@@ -16,3 +19,65 @@ public class ReferencePointSetting(ReferencePointType value) : ICardSetting
     .Cast<ReferencePointType>()
     .ToDictionary(v => v.ToString(), v => v);
 }
+
+public class ReferencePointManager()
+{
+  public Transform? GetTransform(RevitContext context, ReferencePointType referencePointType)
```
oguzhankoral, line 25: You can get `RevitContext` from DI instead passing from one to another

oguzhankoral (changes requested):

oguzhankoral (dismissed):
Once it is tested with proper models. Good to go!

didimitrie (commented):
https://github.com/specklesystems/speckle-sharp-connectors/blob/e821e93ba5e17bea5eaf222f7f8ef04c0b36f42a/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/Raw/Geometry/MeshConversionToSpeckle.cs#L49

Semi unrelated to this PR, but finally found the discussion point from today's standup! We should probably do math on doubles here as @JR-Morgan was saying...

didimitrie (approved):

PR #204, opened 2024-08-27
Fix(revit): Reference to CefSharp.Wpf.NETCore for Revit 2025

oguzhankoral:
TL;DR > Moved from WebView2 to CefSharp.Wpf.NETCore for Revit 2025

Reviewer should test too!
- Open Revit 2025
- Open relatively big model
- Send model
- TRY to navigate viewer of Revit -> Expectation is not frozen Revit UI and responsive UI interactions

Since the thread management is different between WebView2 and CefSharp, Revit UI was freezing in 2025 bc we moved to WebView2 with it. The source of the problem was we were running SendOperation with `RevitTask.RunAsync` to prevent random `AccessMemoryVialoation` exceptions. And running the operation with `RevitTask.RunAsync` was freezing Revit UI with WebView2.

Thanks to @JR-Morgan and @didimitrie for live fixing effort!


didimitrie (approved):
functionally tested, LGTM

PR #205, opened 2024-08-28
ACAD: receive points as fallback displayValue

KatKatKateryna:
This allows to receive objects that have display Value as Points. E.g. Point features from ArcGIS


oguzhankoral (approved):
LGTM

PR #209, opened 2024-08-29
fix update changed IDs

KatKatKateryna:
Fixing the below 2 issues, and moving layerOrder on send also to MapMember utils

Iterating through the map layers was:
1. Incomplete: Map.Layers would only return the top-level Layers, and not check the nested groups
2. Throw if some layers have invalid data source, even if not related to Speckle Send: 
image


Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Utils/MapMembersUtils.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+  public List<MapMember> GetAllMapMembers(Map map)
+  {
+    // first get all map layers
+    Dictionary<MapMember, int> membersIndices = new();
+    int count = 0;
+    var layers = map.Layers;
+    count = UnpackMapLayers(membersIndices, layers, count);
+
+    return membersIndices.Select(x => x.Key).ToList();
```
oguzhankoral, line 17: I'm confused. We are not returning count info, and we are not using the count info as condition then why we track count at all?

oguzhankoral (approved):

PR #215, opened 2024-09-02
Change conversion stack to better usage for products.

adamhathcock:
https://linear.app/speckle/issue/CXPLA-39/refactor-conversioncontextstack-for-greater-flexibility

Use generic conversion stack.
Have records for product specific settings.
Use `with` syntax to push stack values


AlanRynne (approved):
Codewise looks good.

Tested in Rhino, Revit and ACAD and both send/receive with no registration issues.

PR #219, opened 2024-09-04
fix(Revit):CNX-414 fixes DirectShape category bug

kekesidavid:
DirectShapes have a category property of type RevitCategory enum. We are assigning category to all converterd revit elements as a string after conversion. This was resulting in an exception for DirectShapes due to type mismatch. This is a POC fix to skip DirectShapes during category assigment. We should revisit DirectShape class to consider changing category type.

### DIFF
--- a/Converters/Revit/Speckle.Converters.RevitShared/RevitRootToSpeckleConverter.cs
+++ b/Converters/Revit/Speckle.Converters.RevitShared/RevitRootToSpeckleConverter.cs
@@ -48,6 +48,7 @@ public class RevitRootToSpeckleConverter : IRootToSpeckleConverter
 
       // POC DirectShapes have RevitCategory enum as the type or the category property, DS category property is already set in the converter
       // trying to set the category as a string will throw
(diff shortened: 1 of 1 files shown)


clairekuang (approved):

clairekuang:
can we add a POC comment that the category should be moved to be set in each converter instead of the root to speckle converter. We'll probably tackle this when we get around to implementing an IRevitElement interface marking the `category` property as required.

JR-Morgan:
No objections, since this PR simply fixes the current pattern.
However, I would like us to seriously consider us setting the category from the converters themselves rather than doing it here. (same for material quantities)

Yes, this will mean an extra call in every converter, but it means we can use the `required` keyword to make it impossible to forget to attach a category property, and gives us the type safety so a problem like the one this PR addresses can never happen in the first place.

JR-Morgan (dismissed):

didimitrie (approved):

JR-Morgan (approved):

PR #159, opened 2024-08-15
Required keyword on various object models

JR-Morgan:
Connectors half of https://github.com/specklesystems/speckle-sharp-sdk/pull/76

PR #223, opened 2024-09-06
Fix(dui3): Use Invoke function instead InvokeAsync

oguzhankoral:
Reverting the mistake that caused important regression on autocad. TESTED, see gif

acad_2PMJkxaSBS


JR-Morgan (approved):
This change wasn't an intentional inclusion in https://github.com/specklesystems/speckle-sharp-connectors/pull/159

Feel free to merge, but I am suspicious that this part of the code has potential for deadlocks and race conditions.
Perhaps we can take a look after the next beta

PR #224, opened 2024-09-06
Agis fix required units

KatKatKateryna:
fix units setting in ArcGIS


KatKatKateryna:
@JR-Morgan this seems ready to merge, but it hasn't been tested, as Local setup build fails

JR-Morgan (approved):

PR #228, opened 2024-09-09
Switch tracing to seq dev.  Turn off tracing to console when debugging

adamhathcock:


AlanRynne (changes requested):
I think this should go into `main` right? Otherwise everything is good

adamhathcock:
I guess it should go to both.  I'll put it on main so we can put it in dev

adamhathcock:
Actually, this class is newer than the current main :)

Main is like 2 months old, that doesn't seem right 🤔

AlanRynne (approved):
Oh yeah, we still need to do the merge to main. Approving into dev for now, it'll make it's way there

PR #233, opened 2024-09-11
Fix(dui3): Go with default CancellationTokenSource.None for ActionBlock

oguzhankoral:
Somehow, TimeSpan.FromDays(10000) is not good for this. It crashes...... So by default there is no cancellation logic on ActionBlock. I just removed it. @didimitrie let me know if there was a reason adding this CancellationTokenSource..

image


adamhathcock (approved):

didimitrie (approved):

JR-Morgan:
@oguzhankoral are you sure it was an `OperationCancelledException` is the problem? the screenshot shows something about binding initilization.

oguzhankoral:
@JR-Morgan yes it is binding initialization problem and Idk why, you can reproduce it easily, try to run Rhino with previous state

PR #240, opened 2024-09-13
Asyicify `ExecuteScriptAsync` calls

JR-Morgan:
Our browser calls to `ExecuteScriptAsync` were not being awaited
This left room for:
- Race conditions where there were high frequency calls to `ExecuteScriptAsync` (like we saw when we accidently where updating send filters every layer change)
- Exceptions going unobserved
- Inconsistent blocking behaviour between apps that make progress calls on the main thread vs ones that are making progress calls from worker threads.

---

This PR replaces the blocking `Invoke` call with the non-blocking `InvokeAsync` call, and propagates `await async` up to all usages of the browser.
This means we'll need to ensure that we are `awaiting` and using async in a few more areas, such as model card commands.
image

Additionally, to help with the increased usage of `async`, I've changed the `IdleManager`s to have an overload that accepts a `Func<Task>`, added an async version of `RunOnMainThread`, and I've introduced a `FireAndForget` function in the `TopLevelExceptionHandler` for "fire and forget" style event handlers.

The advice going forward:
- We need to be `awaiting` all `Tasks` there are very very very few exceptions to this rule

---

TODO Before ready to review:
 - [x] Test Revit
 - [x] Test Rhino
 - [x] Test Autocad
 - [x] Test ArcGIS
 - [x] Fix tests
 - [x] Get Adam's eyes on this 
 - [ ] Update Operations.Receive & Operations.Send progress


JR-Morgan:
@adamhathcock I'd like to go through this PR with you first, just to make sure my `FireAndForget` logic is sound, and incase you can spot any further improvements.

Once you're happy with it, I'll ask @oguzhankoral and/or @didimitrie for a final review, just to keep them in the loop as to why these changes are important. (the main selling factor is, if we use this `await async` everywhere, we don't have to think about our thread blocking model nearly as much, it just works!)

adamhathcock (approved):
LGTM...only one async void!

oguzhankoral (commented):
I'm triggered to hard test it, if it works as expected, not against it. Just you need to confirm it with @didimitrie too since it touches many critical parts.

PS: Need to also refresh my mind what was the issue once we used `InvokeAsync` instead `Invoke` on Browser.Dispatcher

Async-Await

oguzhankoral (changes requested):
Using `InvokeAsync` on `Browser.Dispatcher` causes an issue on progress update. Please compare your feature branch with `dev` to see difference. If we use `InvokeAsync` progress stuck at the `Starting to send..` step even if it converts

JR-Morgan:
Yep, can reproduce the issue in AutoCAD. Looks the ProgressOperation is not yielding the main thread, so progress is blocked.

I can think of two ways to fix
1. Asyncify the `ProgressOperation
2. Avoid the dispatcher for calls already on main thread (easier to do, but this may not fully resolve)

AlanRynne:
Commenting for posterity:

We found Revit 2023 receive was not working, and the initial approach resulted in a deadlock. Adam is investigating

adamhathcock:
Latest changes:
Revit works with RevitTask and not using default Progress (which no longer syncs to UI thread)
Autocad works with Task on Main thread and Task.Delay(10) which allows UI thread items to process and not using default Progress (which no longer syncs to UI thread)
Rhino doesn't seem to care.

AlanRynne (approved):

PR #244, opened 2024-09-16
Feat(revit) CNX-442 handle instances blocks local to global

oguzhankoral:
It just worked like a charm. Because we did heavy work before for ArcGIS.

image


Sdk/Speckle.Converters.Common.DependencyInjection/ContainerRegistration.cs
```diff
@@ -23,6 +23,7 @@ This will require consolidating across other connectors.
     builder.AddScoped<IFactory<IToHostTopLevelConverter>, Factory<IToHostTopLevelConverter>>();
     builder.AddScoped<IConverterResolver<IToHostTopLevelConverter>, ConverterResolver<IToHostTopLevelConverter>>();
 
+    builder.AddScoped<LocalToGlobalConverterUtils>();
```
didimitrie, line 26: Could we revert back to injecting this just in connectors that need it, rather than in all?

didimitrie (changes requested):
one minor change requested :)

Sdk/Speckle.Converters.Common.DependencyInjection/ContainerRegistration.cs
oguzhankoral: makes sense

oguzhankoral: done

didimitrie (approved):

PR #245, opened 2024-09-17
purge materials and groups in Revit before update

kekesidavid:
purge materials and groups in Revit before update


Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Receive/RevitHostObjectBuilder.cs
```diff
@@ -176,6 +194,12 @@ CancellationToken cancellationToken
     return new(bakedObjectIds, conversionResults);
   }
 
+  private void PreReceiveDeepClean(string baseGroupName)
+  {
+    _groupManager.PurgeGroups(baseGroupName);
```
oguzhankoral, line 199: I would rename `_groupManager` to `_groupBaker` to make it aligned with class name

oguzhankoral (changes requested):
Minor

oguzhankoral (approved):
Tested, works nice!

PR #254, opened 2024-09-18
Revit receive first pass: reference geometry workflow (CNX-403)

didimitrie:
First pass at revit receive, where everything becomes a nested by collection direct shape. 

[CNX-403: Revit Receive: Reference Geometry Workflow](https://linear.app/speckle/issue/CNX-403/revit-receive-reference-geometry-workflow)


oguzhankoral (approved):
Tested,
- Receive from Revit (sample house)
- Receive materials from rhino (by layer, by object)
- Receive blocks and all proxies

PR #256, opened 2024-09-19
Simplify DI assembly structure

adamhathcock:
Remove Autofac dependency.
Removing extra DI and assembles in favor of direct references.
Reworked some dependencies
S.Connectors.Utils -> S.Connectors.Common

[x] Rhino
[x] Revit
[ ] ArcGIS
[x] Autocad

also https://linear.app/speckle/issue/CXPLA-12/autofac-registration-syntax


AlanRynne (approved):
This is looking good. Also tested in autocad with no errors :)

PR #262, opened 2024-09-19
Chore(revit): Refactor material cache to ToHost ToSpeckle singletons

oguzhankoral:
We have now 2 singleton classes ^_____^
I couldn't go further to simplify it more, obviously Revit is different. Materials need to go back and forth between connector and converter.

PS: Refactored (splitted) `MaterialConversionToSpeckle` class into 2. Because it used in 2 different places and one of the return value was always useless, there was unnecessary conversion always.

@adamhathcock not sure is there a way to register an instance in `RevitHostObjectBuilder` that will be used only in that function (also in converter) then it will disposed instead using singleton and clearing the dictionary. Let me know if you have elegant way.


Converters/Revit/Speckle.Converters.RevitShared/Helpers/RevitToSpeckleCacheSingleton.cs
```diff
@@ -1,36 +1,38 @@
...
-/// <para>Persistent cache (across conversions) for all generated render material proxies. This cache stores a list of render material proxies per element id, and provides a method to get out the merged render material proxy list from a set of object ids for setting on the root commit object post conversion phase.</para>
 /// <para>
-/// Why is this needed? Because two reasons: send caching bypasses converter and revit conversions typically generate multiple display values per element. Ask dim for more and he might start crying.
+/// Lifetime of this singleton should be per document.
+/// </para>
+/// <para>
+/// Persistent cache (across conversions) for all generated render material proxies.
+/// This cache stores a list of render material proxies per element id, and provides a method to get out the merged render material proxy list from a set of object ids for setting on the root commit object post conversion phase.
+/// </para>
+/// <para>
```
clairekuang, line 14: my mistake, this para can be deleted

PR #264, opened 2024-09-20
ArcGIS GeodatabaseTableException

JR-Morgan:
Where possible, we should be specific about the exceptions we intend to handle

### DIFF
--- a/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Utils/FeatureClassUtils.cs
+++ b/Converters/ArcGIS/Speckle.Converters.ArcGIS3/Utils/FeatureClassUtils.cs
@@ -44,7 +44,7 @@ public class FeatureClassUtils : IFeatureClassUtils
       schemaBuilder.Build();
       MapView.Active.Redraw(true);
     }
-    catch (GeodatabaseTableException ex)
+    catch (GeodatabaseTableException)
     {
       // "The table was not found." | System.InvalidCast
       // delete Table if already exists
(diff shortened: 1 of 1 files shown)


Converters/ArcGIS/Speckle.Converters.ArcGIS3/Utils/FeatureClassUtils.cs
```diff
@@ -44,7 +44,7 @@ public void ClearExistingDataset(string featureClassName)
       schemaBuilder.Build();
       MapView.Active.Redraw(true);
     }
-    catch (Exception ex) when (!ex.IsFatal()) //(GeodatabaseTableException)
+    catch (GeodatabaseTableException ex)
     {
       // "The table was not found." | System.InvalidCast
```
JR-Morgan, line 49: should this also be handling `System.InvalidCastException`?

KatKatKateryna: looks legit with your changes already, we don't seems to cast anything here, no?

JR-Morgan: Doesn't look to be. But there was this comment so I thought I'd ask.

Perhaps `GetDefinition<FeatureClassDefinition>` may throw if there is something in the db with the name, but it isn't a `FeatureClassDefinition`

Not sure if that's even a possibility.

JR-Morgan:
stale

PR #276, opened 2024-09-24
fix(revit): receives circles

clairekuang:
Revit wasn't receiving circles because the `CircleToHost` converter was producing an **unbounded** arc, and this was throwing when trying to append to a direct shape.

Fixes by bounding the circle in the CircleToHost converter.

Also includes a refactor of the ToHost conversions:
- Removes a bunch of unused top level converters
- clarifies logic for direct shape conversion: (1) converts all objects or object display values into a list of `Revit.GeometryObject` (2) uses converted `Revit.GeometryObject` to create DS. Encapsulates (1) in a new `ITypedConverter<Base, List<DB.GeometryObject>> BaseToHostGeometryConverter` which is injected into `RevitRootToHostConverter`. Previously everything was being done in the `RevitRootToHostConverter`.

image


didimitrie (approved):
tested, works nicely and big thankies for the -247 lines of refactor!

PR #280, opened 2024-09-25
chore(revit): adds units to object params

clairekuang:
Pulls the units from the parameter definition to add to object parameters. If no units are present, there will be no unit field on the object param.

https://latest.speckle.systems/projects/3f895e614f/models/7367fe3688@03e237b76d
image


oguzhankoral (approved):
Code-wise ok to me, but can't say (yet) anything about expected behavior for this.

didimitrie (changes requested):
@clairekuang could we do a very fast performance benchmark on this before we merge? 
basically delete the data.db from appdata/speckle, and send the advanced sample model (should be enough) with and without units on each parameter.

it should be fine overall, but i'd want to quantify what we're sacrificing in terms of performance! 

code wise everything ok!

clairekuang:
> @clairekuang could we do a very fast performance benchmark on this before we merge? basically delete the data.db from appdata/speckle, and send the advanced sample model (should be enough) with and without units on each parameter.
> 

Tested on the sample structural model, adding units was about a 5.6% increase in size :0
image
commit with units: https://latest.speckle.systems/projects/3f895e614f/models/7367fe3688@e30926e581

didimitrie (approved):

PR #283, opened 2024-09-26
fix: Removes some properties from the .targets file and puts them into each csproj

AlanRynne:
> repetition but also much > clarity on what's going on.


oguzhankoral (approved):
TY

PR #284, opened 2024-09-26
CXPLA-89: Replace AutomaticallyUseReferenceAssemblyPackages with nuget

AlanRynne:
⚠️ This is a test to confirm with other team members that the lock files are stable this way.

Ideal scenario would be that no changes to any lock file happen locally in any machine from this commit onwards.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -38,6 +38,6 @@
     <GlobalPackageReference Include="PolySharp" Version="1.14.1" />
     <GlobalPackageReference Include="Microsoft.SourceLink.GitHub" Version="8.0.0" />
     <GlobalPackageReference Include="Speckle.InterfaceGenerator" Version="0.9.6" />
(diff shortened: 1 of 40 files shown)


Directory.Packages.props
```diff
@@ -38,5 +38,6 @@
     <GlobalPackageReference Include="PolySharp" Version="1.14.1" />
     <GlobalPackageReference Include="Microsoft.SourceLink.GitHub" Version="8.0.0" />
     <GlobalPackageReference Include="Speckle.InterfaceGenerator" Version="0.9.6" />
+    <GlobalPackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3" PrivateAssets="All" />
```
adamhathcock, line 41: Don't need `PrivateAssets` since you're using `GlobalPackageReference` https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management#global-package-references

adamhathcock (changes requested):

adamhathcock (approved):

PR #285, opened 2024-09-26
Clean up attributes for logs/traces and user id correlation

adamhathcock:


Sdk/Speckle.Connectors.Common/Connector.cs
```diff
@@ -24,8 +24,8 @@ HostAppVersion version
     TypeLoader.Initialize(typeof(Base).Assembly, typeof(Point).Assembly);
 
     var (logging, tracing) = Observability.Initialize(
+      application.Name + " " + HostApplications.GetVersion(version),
```
oguzhankoral, line 27: changing the order of properties was intentional or not?

adamhathcock: Intentional in that I just put would I thought was most significant to least.  But I did go through some refactorings

Sdk/Speckle.Connectors.Logging/Internal/LogBuilder.cs
```diff
@@ -18,16 +15,9 @@ public static LoggerProvider Initialize(
     ResourceBuilder resourceBuilder
   )
   {
-    var fileVersionInfo = GetFileVersionInfo();
     var serilogLogConfiguration = new LoggerConfiguration()
       .MinimumLevel.Is(SpeckleLogLevelUtility.GetLevel(speckleLogging?.MinimumLevel ?? SpeckleLogLevel.Warning))
       .Enrich.FromLogContext()
-      .Enrich.WithProperty("version", fileVersionInfo.FileVersion)
-      .Enrich.WithProperty("productVersion", connectorVersion)
-      .Enrich.WithProperty("hostOs", DetermineHostOsSlug())
```
oguzhankoral, line 27: hmm, why don't we?

adamhathcock: this is now part of tags and will attach to all traces and logs as part of Otel

AlanRynne (approved):

PR #287, opened 2024-09-27
Jedd/cnx 590 resolve raceconditions where event handlers use the browser

JR-Morgan:
- [x] Fix tests
- [x] Test a little more in connectors


JR-Morgan:
Adam has suggested that an Event Aggregator pattern may also be a potential solution here.
I think this could be cleaner, so I'll give it a POC.

JR-Morgan:
Closing in favour of a EventAggregator Pattern

PR #293, opened 2024-10-09
ArcGIS GroupLayers

KatKatKateryna:
Treat all Collection-alike layers as GroupLayers (converting to Collections). Otherwise, they are passed to the Converter and justifiably failing.

### DIFF
--- a/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/HostApp/ArcGISColorManager.cs
+++ b/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/HostApp/ArcGISColorManager.cs
@@ -502,6 +502,11 @@ public class ArcGISColorManager
         // set the group color to class symbol color if conditions are met
         if (groupConditionsMet)
         {
+          if (groupClass.Symbol is null)
+          {
+            color = RbgToInt(255, 255, 255, 255);
+            return false;
+          }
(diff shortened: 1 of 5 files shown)


Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/HostApp/ArcGISColorManager.cs
```diff
@@ -316,6 +316,10 @@ private void ProcessFeatureLayerColors(FeatureLayer layer, int displayPriority)
     int count = 1;
     using (RowCursor rowCursor = layer.Search())
     {
+      if (rowCursor is null)
+      {
+        return;
+      }
```
AlanRynne, line 322: Does this mean that we can skip an item in a layer if it's null?

Could this happen for a valid layer with valid fields?

Maybe I'm understanding this wrong 😅 but if this is the case we should at least log it somehow.

AlanRynne (approved):

Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/HostApp/ArcGISColorManager.cs
KatKatKateryna: added a comment 🙌

PR #300, opened 2024-10-11
Dimitrie/cnx 510 revit parameter exports perfection

didimitrie:
Fixes a bunch of tickets from revit parameters.


clairekuang (approved):
Generally looks fine to me. Approving, but the `object` type for the param name dictionary took a bit to understand: passing a tuple in the case of `ElementType` to be parsed later - more comments wouldn't hurt here

PR #301, opened 2024-10-11
feat(civil3d): adds all properties for sites, catchments, surfaces, and network parts

clairekuang:
Adds all class and general properties for sites, catchments, surfaces, and network parts.
Also includes CatchmentGroups as `GroupProxies` of any catchment sent, and `Network` as `GroupProxy` containing part ids.


adamhathcock:
Some of these look like caches as they're made using a dictionary and are scoped.  I would probably name them that way.

adamhathcock (approved):

PR #305, opened 2024-10-15
feat(civil3d): adds subassembly and parcel props

clairekuang:
Small pr to add some custom props for subassemblies and parcels


Converters/Civil3d/Speckle.Converters.Civil3dShared/ToSpeckle/BuiltElements/CivilEntityToSpeckleTopLevelConverter.cs
```diff
@@ -35,11 +36,18 @@ CorridorHandler corridorHandler
 
   public Base Convert(CDB.Entity target)
   {
+    string name = target.DisplayName;
+    try
+    {
+      name = target.Name; // this will throw for some entities like labels
+    }
+    catch (Exception e) when (!e.IsFatal()) { }
```
AlanRynne, line 44: Do we know the specific exception? Is there an Autocad generic exception we could use at least?

AlanRynne (approved):

PR #306, opened 2024-10-17
feat(civil3d): adds 2022, 2023, and 2025 versions

clairekuang:
Adds additional versions for civil3d


Connectors/Autocad/Speckle.Connectors.AutocadShared/Plugin/AppUtils.cs
```diff
@@ -10,21 +10,19 @@ public static class AppUtils
 #elif AUTOCAD
     HostApplications.AutoCAD;
 #else
-    throw new NotImplementedException();
+    throw new NotSupportedException();
```
JR-Morgan, line 13: Just a comment,
I think we could use the preprocessor error 
```
#error host application not supported
```
instead, it might give us quicker feedback when we come to add a new project verison

AlanRynne (approved):

adamhathcock (approved):

PR #309, opened 2024-10-18
Dimitrie/cnx 635 curtain wall doors duplicated geometry

didimitrie:
Fixes [CNX-656: Revit: deleted elements cause send and preview selection to fail](https://linear.app/speckle/issue/CNX-656/revit-deleted-elements-cause-send-and-preview-selection-to-fail) and [CNX-635: Curtain wall doors - duplicated geometry](https://linear.app/speckle/issue/CNX-635/curtain-wall-doors-duplicated-geometry)


oguzhankoral (approved):
Will have a look that note you dropped

PR #314, opened 2024-10-21
chore(civil3d): uses preprocessor error instead of throwing exception for unsupported versions

clairekuang:
From @JR-Morgan 's suggestion on this pr: https://github.com/specklesystems/speckle-sharp-connectors/pull/306


clairekuang:
Actually, closing without merging, this behavi checked across all connectors in a separate issue

PR #316, opened 2024-10-21
Fix(revit): CNX-657 revit expiration checks seem to be not fully working for

oguzhankoral:
TLDR; We still use `UniqueId`s.

Main problem was event of deleted objects. We were trying to get `UniqueId` by using provided `ElementId`, but `doc.GetElement(elementId)` can't get any element from deleted `ElementId`. We have to keep remember corresponding `UniqueId`s for `ElementId`s whenever they initially used.


clairekuang (approved):
Only comment: we should check for selection filter behavior when a user deletes a saved object without updating the filter before sending.

PR #322, opened 2024-10-23
Feat(revit): view filter

oguzhankoral:
chrome_gUJwkxj2b4


clairekuang (changes requested):
Let's try a different version of this by prompting users to select the views from the project browser, and treating the UX/UI the same as the selection filter.

Main reasoning behind this is so we don't need new UI components for View filters, and we don't need to worry about keeping the DUI synced with the hostApp state when users add/delete views while in the view filter view.

clairekuang (changes requested):
Let's try a different version of this by prompting users to select the views from the project browser, and treating the UX/UI the same as the selection filter.

Main reasoning behind this is so we don't need new UI components for View filters, and we don't need to worry about keeping the DUI synced with the hostApp state when users add/delete views while in the view filter view.

clairekuang (approved):

Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/Filters/RevitEverythingFilter.cs
didimitrie, line 1: Shall we remove this one if it's not used?

clairekuang:
Only comment on this is that iterating with `GetElement()` through all `ChangedObjIds` just to test for a changed view seems kinda expensive. Might want to compare with using the filtered element collector for doc views, esp on large models

Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/Filters/RevitViewsFilter.cs
```diff
@@ -0,0 +1,78 @@
...
+    var views = collector
+      .OfClass(typeof(View))
+      .Cast<View>()
+      .Where(v => !v.IsTemplate)
+      .Select(v => v.ViewType.ToString() + " - " + v.Name.ToString())
+      .ToList();
+    AvailableViews = views;
+  }
+
+  public void SetContext(RevitContext revitContext)
```
didimitrie, line 73: please add a comment on why we're doing this! (deserialisation, etc.). tl;dr: we are mixing data and behaviour, but i'm not sure how to resolve this pragmatically - i'd poke @adamhathcock during the IRL for some extra wisdom. this is not a blocker for merge though

didimitrie (changes requested):

Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/Filters/RevitEverythingFilter.cs
oguzhankoral: yes, i will

Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/Filters/RevitViewsFilter.cs
oguzhankoral: done

didimitrie (approved):

PR #326, opened 2024-10-24
fix(civil3d/rhino): adds correct curves to displayvalue and receiving arcs as fallback in rhino

clairekuang:
Fixes issue in receiving Civil3d elements with curves in their display value in rhino.
This was done by: 
- prefiltering list of ICurve in Civil3d to only include lines, polylines, and arcs to be used as display values.
- adding arcs to the supported types of fallback geometry in rhino

sample commit: https://latest.speckle.systems/projects/3f895e614f/models/1737754fb4@1d2b032ba5

received in rhino:
{589AE953-792B-480C-B158-47E66C07B06B}

extra check received in autocad:
{3C905819-DC10-4CBA-872D-F0D7FDB23134}


adamhathcock (approved):
recursive = true 😬

PR #327, opened 2024-10-25
feat(rhino): adds names to all objects on bake

clairekuang:
Looks for a `name` property on incoming atomic objects, and bakes rhino objects with a name if one exists.
Motivation: while the name is nice to preserve for multiplayer flows, this property helps a lot when receiving fallback geometry from other apps (esp civil and revit) that would help identify what the original object was.

Also adds `name` property to all rhino objects on send.

### DIFF
--- a/Connectors/Rhino/Speckle.Connectors.RhinoShared/Operations/Receive/RhinoHostObjectBuilder.cs
+++ b/Connectors/Rhino/Speckle.Connectors.RhinoShared/Operations/Receive/RhinoHostObjectBuilder.cs
@@ -129,6 +129,9 @@ public class RhinoHostObjectBuilder : IHostObjectBuilder
           onOperationProgressed.Report(new("Converting objects", (double)++count / atomicObjects.Count));
           try
           {
+            // 0: get the name of the incoming obj if any
+            string? name = obj["name"] as string;
+
             // 1: get pre-created layer from cache in layer baker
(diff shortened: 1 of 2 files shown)


Connectors/Rhino/Speckle.Connectors.RhinoShared/Operations/Receive/RhinoHostObjectBuilder.cs
```diff
@@ -253,7 +256,7 @@ private void PreReceiveDeepClean(string baseLayerName)
     _groupBaker.PurgeGroups(baseLayerName);
   }
 
-  private Guid BakeObject(GeometryBase obj, Base originalObject, int layerIndex)
+  private Guid BakeObject(GeometryBase obj, Base originalObject, int layerIndex, string? name = null)
```
oguzhankoral: since we already have `originalObject` in relevant functions, I wouldn't pollute function arguments. You need to do it only in BakeObjects function and thats it

oguzhankoral (changes requested):

Connectors/Rhino/Speckle.Connectors.RhinoShared/Operations/Receive/RhinoHostObjectBuilder.cs
clairekuang: no I can't do it there, since the original object could be a line in a display value of a civil element for example

oguzhankoral (approved):

PR #333, opened 2024-10-30
refactor(objects): cnx 687 purge unused classes from objects

clairekuang:
Changes to all converters pending this pr in sdk: https://github.com/specklesystems/speckle-sharp-sdk/pull/151 . To be merged after SDK pr is merged.

- removes qgis backwards compatibility and revit/rhino parameters and user strings in arcgis 
- removes unused parameter, material, and material quantities code in revit
- refactors revit `modelcurves` by sending them as regular geometry instead


Converters/ArcGIS/Speckle.Converters.ArcGIS3/Utils/ArcGISFieldUtils.cs
```diff
@@ -186,49 +186,10 @@ List<string> fieldAdded
...
-        }
-      }
-      else
-      {
-        // for now, ignore all other properties of Base type
-      }
+      // Revit parameters are sent under the `properties` field as a `Dictionary<string,object?>`.
+      // This is the same for attributes from other applications. No Speckle objects should have attributes of type `Base`.
+      // Currently we are not sending any rhino user strings.
+      // TODO: add support for attributes of type `Dictionary<string,object?>`
```
KatKatKateryna, line 192: we can just move this part of code lower to the condition "if Dict" rather than if Base. Why removing if the TODO is already done here?

KatKatKateryna (changes requested):
Checked ArcGIS part - looks good except the Parameters converter, it will look nearly the same for Dictionary object and can be reused for that.

Converters/ArcGIS/Speckle.Converters.ArcGIS3/Utils/ArcGISFieldUtils.cs
clairekuang: Let's sync on this: archicad, civil3d, revit, and eventually tekla will send some properties/parameters under the `properties` field with a dict value - need to implement this in a way in arcgis that can handle diff levels of nesting

adamhathcock (approved):
Works for me

JR-Morgan (approved):

KatKatKateryna (approved):

PR #334, opened 2024-10-30
Dogukan/cnx 690 create top level converters for all tekla objects

dogukankaratas:
adds top level converter and property extractors


clairekuang (changes requested):

Converters/Tekla/Speckle.Converter.Tekla2024/ToSpeckle/Helpers/DisplayValueExtractor.cs
```diff
@@ -0,0 +1,48 @@
...
+    _meshConverter = meshConverter;
+    _settingsStore = settingsStore;
+  }
+
+  public IEnumerable<Base> GetDisplayValue(TSM.ModelObject modelObject)
+  {
+    switch (modelObject)
+    {
+      case TSM.Beam beam:
+        var solid = beam.GetSolid();
```
clairekuang, line 27: Is `GetSolid()` on the `Part` level? If so we can just have one `TSM.Part` case to retrieve the solid from

Converters/Tekla/Speckle.Converter.Tekla2024/ToSpeckle/Raw/LineToSpeckleConverter.cs
clairekuang, line 1: Let's keep this converter

clairekuang (changes requested):

clairekuang (approved):

PR #344, opened 2024-11-03
unpack parameters as Dictionary

KatKatKateryna:
Unpacking parameters assuming they are Dictionary rather than a Base. Works with simple attributes, but Revit models seems to start sending null in "parameters" (e.g. https://latest.speckle.systems/projects/126cd4b7bb/models/523fc1299b ), need newer model to test


clairekuang:
@KatKatKateryna they are sent under ‘properties’ not ‘parameters’

clairekuang:
Closing, arcgis receive is disabled and this should be handled on dataobjects

PR #352, opened 2024-11-05
Dogukan/cnx 691 extract appropriate class properties by type

dogukankaratas:
attachs class and report properties to the model objects


Converters/Tekla/Speckle.Converter.Tekla2024/ToSpeckle/Helpers/ReportHandler.cs
```diff
@@ -0,0 +1,88 @@
...
+          "ASSEMBLY_POS",
+          "PHASE"
+        }
+      },
+      { typeof(TSM.RebarGroup), new[] { "NUMBER_OF_REBARS", "TOTAL_LENGTH", "WEIGHT", "SIZE", "GRADE", "CLASS" } },
+      { typeof(TSM.SingleRebar), new[] { "LENGTH", "WEIGHT", "SIZE", "GRADE", "CLASS" } },
+      { typeof(TSM.BoltArray), new[] { "BOLT_SIZE", "NUMBER_OF_BOLTS", "BOLT_STANDARD", "BOLT_TYPE", "LENGTH" } }
+    };
+
+  public Dictionary<string, object?> GetProperties(TSM.ModelObject modelObject)
```
oguzhankoral, line 47: if it is in converter, I would rather to follow "converter" logic instead random class and function names if we don't have a strong reason. lacking context shout me if im wrong

oguzhankoral (approved):

PR #368, opened 2024-11-07
fix: Prevent api dependencies from being copied to outDir

AlanRynne:


oguzhankoral:
since it is draft and very old, i am closing it to clean up our PR page. we can resurface if needed with another PR. (if you get pinged Alan, hi! :))

PR #372, opened 2024-11-08
Use Send2 again!

adamhathcock:
Needed if this works out: https://github.com/specklesystems/speckle-sharp-sdk/pull/164


oguzhankoral (approved):
I have no objection on connectors side!

PR #382, opened 2024-11-12
dogukan/tekla2023 installer

dogukankaratas:
some changes to support tekla2023 in installer


Connectors/Tekla/Speckle.Connector.TeklaShared/SpeckleTeklaPanelHost.cs
```diff
@@ -24,10 +24,12 @@ public SpeckleTeklaPanelHost()
     this.Text = "Speckle (Beta)";
     this.Name = "Speckle (Beta)";
 
+    string version = GetVersion().ToString()[1..]; // removes the 'v' from version
```
oguzhankoral, line 27: seems odd. Why we need version to be able get icon, isn't it already living in shared project?

oguzhankoral (changes requested):

Connectors/Tekla/Speckle.Connector.TeklaShared/SpeckleTeklaPanelHost.cs
AlanRynne: If we need the icon here too, we might as well just embed it as an EmbeddedResource accessible through `Properties.Resources.My_BitMap`

PR #388, opened 2024-11-18
Fix(store): lock list modifications on DocumentModelStore

adamhathcock:
Suspend usage of ObservableCollection when doing mass updating.  Only have one and modify it.

CNX-799

https://linear.app/speckle/issue/CNX-799/observablecollection-errors


DUI3/Speckle.Connectors.DUI/Models/DocumentModelStore.cs
```diff
@@ -10,44 +13,17 @@ namespace Speckle.Connectors.DUI.Models;
...
-  public ObservableCollection<ModelCard> Models
-  {
-    get => _models;
-    protected set
-    {
-      _models = value;
-      RegisterWriteOnChangeEvent();
-    }
-  }
+  private readonly List<ModelCard> _models = new();
```
oguzhankoral, line 16: We need to have public `Models` property for UI. Otherwise we cannot serialize `DocumentModelStore` as expected whenever `GetDocumentState()` is called in base bindings.

See the error in console

image

oguzhankoral (changes requested):

adamhathcock:
Can we put things on an interface and/or comments about what is needed for the javascript UI?

oguzhankoral:
> Can we put things on an interface and/or comments about what is needed for the javascript UI?

@adamhathcock 
Initially `DiscriminatedObject` was the idea of this to mark they are data classes for JS. But I guess it is semi-broken now. You are right. We need those interfaces to imply that data/prop is needed for UI.

Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadDocumentModelStore.cs
```diff
@@ -64,6 +63,10 @@ protected override void LoadState()
     }
 
     string? serializedModelCards = _autocadDocumentManager.ReadModelCards(doc);
+    if (serializedModelCards == null)
+    {
+      return;
```
oguzhankoral, line 68: we might need to clear the existing list if we are opening a new empty model, need to test

oguzhankoral (changes requested):

Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadDocumentModelStore.cs
oguzhankoral: yes this is not correct in Autocad. If it is null we must clear the existing list

oguzhankoral (approved):

DUI3/Speckle.Connectors.DUI/Models/DocumentModelStore.cs
```diff
@@ -65,6 +65,10 @@ public void UpdateModel(ModelCard model)
     lock (_models)
     {
       var m = _models.First(m => model.ModelCardId == m.ModelCardId);
+      if (m is null)
```
adamhathcock, line 68: FirstOrDefault

DUI3/Speckle.Connectors.DUI/Models/DocumentModelStore.cs
```diff
@@ -75,7 +79,13 @@ public void RemoveModel(ModelCard model)
   {
     lock (_models)
     {
-      _models.Remove(model);
+      var m = _models.First(m => model.ModelCardId == m.ModelCardId);
+      if (m is null)
```
adamhathcock, line 83: FirstOrDefault

PR #389, opened 2024-11-18
CNX-751: Add Units to reportProperties

bjoernsteinhagen:
## Description & motivation

- Related to [CNX-751](https://linear.app/speckle/issue/CNX-751/add-units-to-reportproperties)

## Changes:

- Use of `Tekla.Structures.Datatype` `enums` for units instead of `Tekla.Structures.Drawing`. This makes more sense as the `Drawing` class is more geared to dimension lines, annotations etc. than the model. Here, we also get rid of the `FeetsAndInches`, `CentimetersOrMeters` and `Automatic` enums.
- `hostUnit` from `Tekla.Structures.Datatype.Distance.CurrentUnitType` returns the correct unit set in `Settings > Options > Units and decimals`, but API returns are always mm. Potential mismatch in model units and report property BUT we append the units to each report property.
- `reportProperties` are default mm (or power of). Hence, units appended are hard-coded. Awaiting response and assistance from Tekla. This can be done much better.

## Screenshots:

image

- tl;dr : value extraction and assignment correct. Hard-coded to mm and kg as this is Tekla internal.

More explanation:
- Tekla API has methods to convert internal units to model units (`Settings > Options > Units and decimals`). But these methods only apply to `Distance` and `Angle` params. Probably more risky to go through series of conversions AND writing additional ones for area, volume etc. than for us to just report everything as internal units.
- `AREA` is not the cross-sectional area of the section but some other value. Not our issue though, this is as per Tekla report
- `VOLUME` is, however, calculated using the true cross-sectional area multiplied by length. Weird.

## Validation of changes:

- Sending from `Basic Steel Model.db1` and checking returned values with generated report.

## To Do Before Merge

- Test with environment in USA @dogukankaratas? **To test our theory if internal unit is somehow dependant on environment**.

## Checklist:

- [x] My pull request follows the guidelines in the [Contributing guide](https://github.com/specklesystems/speckle-server/blob/main/CONTRIBUTING.md)?
- [x] My pull request does not duplicate any other open [Pull Requests](../../pulls) for the same update/change?
- [x] My code follows a similar style to existing code.

### DIFF
--- a/Connectors/Rhino/Speckle.Connectors.RhinoShared/Bindings/RhinoSendBinding.cs
+++ b/Connectors/Rhino/Speckle.Connectors.RhinoShared/Bindings/RhinoSendBinding.cs
@@ -50,6 +50,8 @@ public sealed class RhinoSendBinding : ISendBinding
   /// </summary>
   private ConcurrentDictionary<string, byte> ChangedObjectIds { get; set; } = new();
 
+  private UnitSystem PreviousUnitSystem { get; set; }
+
   public RhinoSendBinding(
     DocumentModelStore store,
     IAppIdleManager idleManager,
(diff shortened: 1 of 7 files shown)


oguzhankoral (approved):
Looks good to me! And thanks for saving me trying to understand whats up, nice screenshots! ❤️
if @dogukankaratas doesn't have any objection or catch, we can merge it.

dogukankaratas (approved):

PR #391, opened 2024-11-19
Dim/type change cach invalidation

didimitrie:
[CNX-805: Changing an ElementType should flag all objects of that type as changed](https://linear.app/speckle/issue/CNX-805/changing-an-elementtype-should-flag-all-objects-of-that-type-as)

Tested in 2023, 2024 and 2025

https://github.com/user-attachments/assets/906ffcd7-1bd9-481f-ba32-a9b266400864

Tested with the revit performance model as well, feels ok - but we defs need to get the idle manager to call things once asap-ish.


Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
```diff
@@ -346,15 +346,33 @@ private async Task RunExpirationChecks()
     }
 
     var objUniqueIds = new List<string>();
+    var changedIds = ChangedObjectIds.Keys.ToList();
+
+    // Handling type changes: if an element's type is changed, we need to mark as changed all objects that have that type.
+    // Step 1: get any changed types
+    var elementTypeIdsList = changedIds
+      .Select(e => doc.GetElement(e))
+      .Where(el => el is ElementType)
```
adamhathcock, line 355: OfType? https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.oftype?view=net-9.0

Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
```diff
@@ -346,15 +346,33 @@ private async Task RunExpirationChecks()
     }
 
     var objUniqueIds = new List<string>();
+    var changedIds = ChangedObjectIds.Keys.ToList();
```
adamhathcock, line 349: Perf: Do we need to make another collection here?

Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
didimitrie: legit, changing

Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
didimitrie: i am adding to it below, so i kinda need it in there

oguzhankoral (approved):

PR #392, opened 2024-11-19
dogukan/cnx-744-model-card-highlight-in-tekla

dogukankaratas:
adds model card highlight and report object highlight


oguzhankoral (approved):
Looks neat!

PR #398, opened 2024-11-20
dogukan/cnx-739-object-tracking-in-tekla

dogukankaratas:
## Description & motivation

- Adds the functionality to notify user if there is any change in the objects already sent to Speckle.
- Does not trigger the idle for now, will figure out that in the future.
- Works together with the cache to not to convert already converted objects.

## Changes:
- Events.OnModelObjectChange class listens the Tekla instance and catch modified object ID's.

image

- [x] My pull request follows the guidelines in the [Contributing guide](https://github.com/specklesystems/speckle-server/blob/main/CONTRIBUTING.md)?
- [x] My pull request does not duplicate any other open [Pull Requests](../../pulls) for the same update/change?
- [x] My code follows a similar style to existing code.


oguzhankoral (approved):
LGTM. @dogukankaratas could you create follow up ticket for idling issue? Or would be nice if you can investigate how bad the current status without idling. For example test to change maaaany elements at once, move all etc

PR #405, opened 2024-11-22
Rework progress to try to be better 😬 

adamhathcock:


Sdk/Speckle.Connectors.Common/Operations/SendProgress.cs
```diff
@@ -0,0 +1,44 @@
...
+    }
+    if (!progressDisplayManager.ShouldUpdate())
+    {
+      return;
+    }
+
+    switch (args.ProgressEvent)
+    {
+      case ProgressEvent.CachedToLocal:
+        onOperationProgressed.Report(new($"Caching... ({args.Count} total objects)", null));
```
oguzhankoral, line 32: not sure cache reporting is a valueable info for user. I wouldn't report it as progress.

oguzhankoral (changes requested):

oguzhankoral (approved):

PR #408, opened 2024-11-25
receive colors with fallback conversions

KatKatKateryna:
tested with https://latest.speckle.systems/projects/2af60ce1b6/models/67aa9f5d4c@81655b012c
image


clairekuang (changes requested):
Instead of passing the entire object, would be cleaner to pass the applicationId string instead

oguzhankoral (changes requested):

oguzhankoral (approved):

oguzhankoral (approved):

clairekuang (approved):

PR #399, opened 2024-11-20
refactor(revit): aligns revit send for data extraction workflow

clairekuang:
- removes direct conversions of a lot of built elements classes and instead sends most revit elements as `RevitObject`
- cleans up properties on some revit elements, mostly display meshes and hosted elements logic
- adds `profile` property to a lot of classes in the `classpropertyextractor`. This replaces the `outline` and `voids/openings` prop of the builtelelements classes, and resolves old bugs on identifying outline curve vs void curves. 
- adds `location` property by defualt to all objects that have it. previously these were sent as `baseCurve`, `baseLine`, or `position/point` props on structural and instance elements.
- fixes railing bug
- fixes empty stacked wall display value bug
- note: `Levels` are still attached on every object as a `Level` object. should refactor this to use level proxies instead in another pr.

Before:
[https://app.speckle.systems/projects/b53a53697a/models/00daea8ea0@3bb4c0ac17](https://app.speckle.systems/projects/b53a53697a/models/00daea8ea0@3bb4c0ac17
)
{7E025E89-3A0F-4766-ADC0-F86E79D8830A}

After:
[https://app.speckle.systems/projects/b53a53697a/models/00daea8ea0@0a119be678](https://app.speckle.systems/projects/b53a53697a/models/00daea8ea0@0a119be678)
{79A17A5E-70F1-445A-AA02-43A2B7234D50}

PR #410, opened 2024-11-26
Use Send2 and latest SDK

adamhathcock:


adamhathcock:
merged in with https://github.com/specklesystems/speckle-sharp-connectors/pull/399

PR #416, opened 2024-11-28
Arcgis refactor

KatKatKateryna:
Using GisObject for sending anything
Move all layer operations from Converter to Connector -> Layer Manager
Remove all top-level converters

Sample of the "after": https://latest.speckle.systems/projects/2af60ce1b6/models/f537cf39e4@f9fa35ce30
image

Sample of "before": https://latest.speckle.systems/projects/2af60ce1b6/models/f537cf39e4@3e734bf8ea
(also, "before" polygons had a "geometry" field that's now not being sent)
image


clairekuang (changes requested):
Let's do a live pr review to discuss changes. Overall notes are:
- **create a `LayerHandler` class in the connector for all `Collection` creation and management logic**. This should cover all gis layer subtypes, and return the appropriate error if a user has selected an unsupported layer type. We want to avoid extensive if/else blocks.
- **Create a properties extractor and display value extractor class** to be injected into the TopLevel converter that converts `GisObject`s. This structure is the same for civil3d, tekla, etabs, etc.
- the **TopLevelConverter should probably be typed to `CoreObjectsBase`** since that appears to be the parent class of all layer objects, and should throw if an unsupported type of `CoreObjectsBase` is passed *before* getting into conversion logic. This is also to avoid extensive if/else blocks to handle subtypes.

KatKatKateryna:
> Let's do a live pr review to discuss changes. Overall notes are:
> 
> * **create a `LayerHandler` class in the connector for all `Collection` creation and management logic**. This should cover all gis layer subtypes, and return the appropriate error if a user has selected an unsupported layer type. We want to avoid extensive if/else blocks.
> * **Create a properties extractor and display value extractor class** to be injected into the TopLevel converter that converts `GisObject`s. This structure is the same for civil3d, tekla, etabs, etc.
> * the **TopLevelConverter should probably be typed to `CoreObjectsBase`** since that appears to be the parent class of all layer objects, and should throw if an unsupported type of `CoreObjectsBase` is passed _before_ getting into conversion logic. This is also to avoid extensive if/else blocks to handle subtypes.

let's have a chat! some quick points in advance:
- LayerUnpacker is currently handling layer conversion logic
- properties and display extractors cannot be moved to GisFeature converter, because the depend on the layer and its properties (currently the 'attributeConverter' is doing that in RootBuilder)

clairekuang:
@KatKatKateryna fixes for your pr change requests from my merged pr are merged in!

KatKatKateryna:
all seems to be resolved, just 1 color bug remains:
test project: https://drive.google.com/drive/u/0/folders/1xTYt9-BAWD7d_scwpsj4nOas7_MbJhTd
image
https://latest.speckle.systems/projects/2af60ce1b6/models/f537cf39e4@da2c7c5855

KatKatKateryna:
2 problems with new unflattening approach:
- report on Send will display X layers while Selection filter showed Y layers (group layers will be skipped in reporting)
- If user selects ONLY the nested layer, nothing will be sent

KatKatKateryna:
Proposed solution:
- revert RemoveNestedMapMembers commit
- return unflattening layers by order, e.g. which will ensure the layers are converted hierarchically and not depending on the user click order. Also integrate the new UnpackSelectionAsync logic into this: 
image
image

KatKatKateryna:
fixing the layer order & removing duplication seems to have fixed the color proxies too

KatKatKateryna:
Issue with the colors:
The problematic part is that getting Row's applicationId via "row.Handle" call is not resulting is a consistent id, therefore the values added to ColorProxies are not corresponding to the correct objects. 
image
image
image
@clairekuang

KatKatKateryna:
Currently group layers and non-feature layers will be assigned this property, not sure we need it: 
image
image

Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Utils/MapMembersUtils.cs
```diff
@@ -41,20 +41,18 @@ public List<MapMember> UnpackMapLayers(IEnumerable<MapMember> mapMembersToUnpack
   }
 
   // Gets the layer display priority for selected layers
```
clairekuang, line 43: Let's update the comment here akin to:
```cs
/// <summary>
/// Sorts the selected mapmembers into the same order as they appear in the Table of Contents (TOC) bar in the file.
/// This is a required step before unpacking layers, because depending on the user selection order, some children layers may appear before their container layer if both the container and children layers are selected.
/// </summary>
public List<MapMember> GetMapMembersInOrder(Map map, IReadOnlyList<MapMember> selectedMapMembers)
  ```

File: Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Operations/Send/ArcGISRootObjectBuilder.cs [hunk 2]
```diff
@@ -274,9 +285,29 @@ await QueuedTask
...
 
     return convertedObjects;
   }
+
+  /// <summary>
+  /// Retrieves the Speckle application id for Features as a concatenation of the layer URI (applicationId)
+  /// and the row OID (index of row in layer) or point OID for LasDatasets.
+  /// </summary>
+  /// <exception cref="ACD.Exceptions.GeodatabaseException">Throws when this is *not* called on MCT. Use QueuedTask.Run.</exception>
+  public string GetSpeckleApplicationId(ADM.Layer layer, AC.CoreObjectsBase coreObject)
```
clairekuang, line 299: Move this to the SpeckleApplicationId extension method class

clairekuang (approved):
All good!

PR #421, opened 2024-11-28
Dimitrie/cnx 855 rhino changing a material does not invalidate the objects

didimitrie:
See linear tix for details: we now invalidate all objects whose material property has changed, eg. color: 

https://github.com/user-attachments/assets/d048e722-4916-4304-8299-161157fcd509

### DIFF
--- a/Connectors/Rhino/Speckle.Connectors.RhinoShared/Bindings/RhinoSendBinding.cs
+++ b/Connectors/Rhino/Speckle.Connectors.RhinoShared/Bindings/RhinoSendBinding.cs
@@ -155,13 +155,14 @@ public sealed class RhinoSendBinding : ISendBinding
       });
 
     RhinoDoc.MaterialTableEvent += (_, args) =>
-    {
-      if (args.EventType == MaterialTableEventType.Modified)
+      _topLevelExceptionHandler.CatchUnhandled(() =>
       {
-        ChangedMaterialIndexes[args.Index] = 1;
-        _idleManager.SubscribeToIdle(nameof(RhinoSendBinding), RunExpirationChecks);
-      }
-    };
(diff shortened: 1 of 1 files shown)


Connectors/Rhino/Speckle.Connectors.RhinoShared/Bindings/RhinoSendBinding.cs
```diff
@@ -266,11 +279,22 @@ private async Task RunExpirationChecks()
       _logger.LogError("Rhino expiration checks were running without an active doc.");
       return;
     }
-    var senders = _store.GetSenders();
-    string[] objectIdsList = ChangedObjectIds.Keys.ToArray(); // NOTE: could not copy to array happens here
-    List<string> expiredSenderIds = new();
 
+    // Invalidate any objects whose materials have changed
+    var changedMaterialIndexes = ChangedMaterialIndexes.Keys.ToArray();
+    foreach (var rhinoObject in RhinoDoc.ActiveDoc.Objects)
```
oguzhankoral, line 285: i would iterate all objects only if `ChangedMaterialIndexes` has any

oguzhankoral (changes requested):

Connectors/Rhino/Speckle.Connectors.RhinoShared/Bindings/RhinoSendBinding.cs
didimitrie: True, will do tomorrow!

didimitrie: done!

oguzhankoral (approved):

PR #424, opened 2024-11-29
Fix for Autocad receive speed

adamhathcock:
There's no way around having to yield to the UI thread to get a status update.  However, we can reduce the number of times we do so.  This is a quick way to have some updates (factor of 50 reduction) but we could do something more inteligent.

Event binding should be reviewing.  This reduces the allocations of the exception handler tasks.  Needs further work though.

https://linear.app/speckle/issue/CNX-854/autocad-receive-performance-totally-tanked

### DIFF
--- a/Connectors/Autocad/Speckle.Connectors.AutocadShared/Operations/Receive/AutocadHostObjectBuilder.cs
+++ b/Connectors/Autocad/Speckle.Connectors.AutocadShared/Operations/Receive/AutocadHostObjectBuilder.cs
@@ -63,13 +63,15 @@ public class AutocadHostObjectBuilder : IHostObjectBuilder
     // NOTE: This is the only place we apply ISyncToThread across connectors. We need to sync up with main thread here
     //  after GetObject and Deserialization. It is anti-pattern now. Happiness level 3/10 but works.
     return await _syncToThread
-      .RunOnThread(
-        async () => await BuildImpl(rootObject, projectName, modelName, onOperationProgressed).ConfigureAwait(false)
-      )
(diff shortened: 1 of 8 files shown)


Connectors/Autocad/Speckle.Connectors.AutocadShared/Bindings/AutocadSendBaseBinding.cs
```diff
@@ -86,6 +86,10 @@ ISpeckleApplication speckleApplication
     {
       _sendConversionCache.ClearCache();
     };
+    _idleManager.SubscribeToIdle(
```
oguzhankoral, line 89: very confused, why we need to subscribe to idle on constructor? Also isn't this PR receive fix, why we are touching send?

Connectors/Autocad/Speckle.Connectors.AutocadShared/Operations/Receive/AutocadHostObjectBuilder.cs
```diff
@@ -127,10 +127,13 @@ await _materialBaker
     {
       string objectId = atomicObject.applicationId ?? atomicObject.id;
       onOperationProgressed.Report(new("Converting objects", (double)++count / atomicObjects.Count));
+      if (count % 50 == 0)
```
oguzhankoral, line 130: don't we have debouncing already for this?

oguzhankoral (changes requested):

JR-Morgan (approved):

oguzhankoral (approved):

PR #428, opened 2024-12-01
feat(autocad): adds surface and region converters

clairekuang:
- adds support for sending surfaces and regions as meshes
- adds brep to mesh raw converter
- refactors existing solid and body converters to use new brep converter to avoid duplicate logic

before:
https://latest.speckle.systems/projects/25d1713b7f/models/87d7a331a8@031076f7e9#threadId=0912fbdc42

after:
https://latest.speckle.systems/projects/2295cb26a0/models/7c9dc1dcd8@12c41008e6


didimitrie (approved):

didimitrie:
I've tested this with the existing sample file; unsure what other things to look at. Had some initial comments, but nothing really worth dropping in here.

PR #430, opened 2024-12-02
Use Sets and Freezing to make conversions faster

adamhathcock:
replaces a lot of List usage and Contains/AddRange/RemoveRange etc should be faster


JR-Morgan (approved):
Tested in Revit

PR #444, opened 2024-12-11
Add Net8  targets for common and fix revit 2025

adamhathcock:
I was still having Revit 2025 build issues.

Since 25 doesn't depend on webview anymore (just Cef) then remove the dependency and blocker for me


JR-Morgan (approved):

Sdk/Speckle.Connectors.Common/Speckle.Connectors.Common.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
```
oguzhankoral, line 4: why we need multi targeting for common project?

adamhathcock: all users of that project that are net8.0 will use the net8.0 version

PR #445, opened 2024-12-11
Dimitrie/cnx 917 steel proxy elements do not get converted out

didimitrie:
See linear ticket for details. 

NOTE: @bimgeek I'd love to test against a more extensive file with more structural connectors/plates/welds/bolts. I have only tested vs. steel plates for now before we merge. 

Alternatively, we can cut a release and ask the issue reporter on the forum (Andre) to give it a go.


oguzhankoral (approved):
We can cut a release with it since we need to do tmr already. Other fixes need to go in.

bimgeek:
@didimitrie tested and issue seems to persist.

didimitrie:
Yep, i forgot to remove preprocessor directives and i assume you've tested with revit 23 or lower?

PR #447, opened 2024-12-12
Dimitrie/cnx 917 steel proxy elements do not get converted out

didimitrie:
This is a rehash of the last PR, changes are that it's now valid across revit versions and should be (hopefully) exporting correct geometry even when structural elements are joined together. 

Let's not merge until @bimgeek approves via a 2025 test (it's broken on my end) 

image
https://latest.speckle.systems/projects/126cd4b7bb/models/1948809dcb


bimgeek (approved):
Just tested in Revit 2025. Seems to work as expected. 

image

PR #452, opened 2024-12-18
Use Result objects to have less exceptions in converters

adamhathcock:
Add some result objects for when exceptions should be thrown.  This results in a huge speed increase

Messages no longer have a stacktrace though:
image


adamhathcock:
https://github.com/specklesystems/speckle-sharp-connectors/pull/456 superceeds this for now

PR #463, opened 2024-12-24
fix(Navisworks): CNX-973 Fix: Material ID Logic and Render Proxy Cache

jsdbroughton:
This update addresses an issue in the `materialID` logic, where the `renderProxies` material cache was based solely on colour. This caused transparent elements sharing the same colour to incorrectly reuse the same material, leading to rendering errors.

Additionally, the material unpacking routine has been improved to cross-check grouped geometry nodes. Previously, unnamed sibling nodes with no unique properties could result in incorrect overrides or omissions of render proxies for child geometries. The update introduces a new parameter for grouped nodes, ensuring the parent group does not improperly affect the `displayValue` render proxies of its children.

also fixes: CNX-893

### DIFF
--- a/Connectors/Navisworks/Speckle.Connectors.NavisworksShared/Operations/Send/NavisworksRootObjectBuilder.cs
+++ b/Connectors/Navisworks/Speckle.Connectors.NavisworksShared/Operations/Send/NavisworksRootObjectBuilder.cs
@@ -1,4 +1,4 @@
-using Microsoft.Extensions.Logging;
+﻿using Microsoft.Extensions.Logging;
 using Speckle.Connector.Navisworks.HostApp;
 using Speckle.Connector.Navisworks.Services;
 using Speckle.Connectors.Common.Builders;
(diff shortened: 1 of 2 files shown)


Connectors/Navisworks/Speckle.Connectors.NavisworksShared/Operations/Send/NavisworksRootObjectBuilder.cs
```diff
@@ -35,6 +37,10 @@ public RootObjectBuilderResult Build(
     CancellationToken cancellationToken
   )
   {
+#if DEBUG
+    // This is a temporary workaround to disable node merging for debugging purposes - false is default, true is for debugging
+    SkipNodeMerging = false;
```
oguzhankoral, line 42: seems marked as `false` on debugging 🤔 but it has no effect for sure on production

oguzhankoral (approved):

Connectors/Navisworks/Speckle.Connectors.NavisworksShared/Operations/Send/NavisworksRootObjectBuilder.cs
jsdbroughton: This is intentional. It happens to be the last time I ran it I set it as false because I didn't want to test for that. I know it's something that I want to test in the future again this Pragma is literally so I don't accidentally leave it set to true for production. 

Long term I'll remove it.

PR #464, opened 2024-12-24
fix(navisworks): Update Navisworks connector plugin bundle target and version content target paths.

jsdbroughton:
- Update the plugin bundle target path to "$(AppData)\Autodesk\ApplicationPlugins\Speckle.Connectors.Navisworks.bundle".
- Update the plugin version content target path to "$(AppData)\Autodesk\ApplicationPlugins\Speckle.Connectors.Navisworks.bundle\Contents\$(NavisworksVersion)".

These changes ensure that the correct paths are used for the Navisworks connector plugin bundle and its version content.


oguzhankoral (approved):
❤️

PR #467, opened 2025-01-06
Rhino panels should be treated as if they might have been used before…

adamhathcock:
… as Panels.Closed isn't called


didimitrie (commented):
https://github.com/user-attachments/assets/b02433fa-b537-47d8-9041-21e6666f74e2

I can seem to still break it, though it is better than before. If you have an idea on how to make it better, let's go - if not, i'll approve

adamhathcock:
Now only a Rhino 8+ change.

Panels can still be made to be empty like the video but it's better than it was.

oguzhankoral (approved):

didimitrie (approved):

PR #468, opened 2025-01-07
Dimitrie/cnx 938 did not find a layer in cache

didimitrie:
See linear ticket and extra comments in there: https://linear.app/speckle/issue/CNX-938/did-not-find-a-layer-in-cache#comment-a1ed59b5


Converters/Rhino/Speckle.Converters.RhinoShared/ToHost/TopLevel/DisplayableObjectToHostTopLevelConverter.cs
```diff
@@ -0,0 +1,77 @@
+using Speckle.Converters.Common;
```
didimitrie, line 1: Note, this should not be needed once archicad sends data objects; but until then/for future bases with display values, this is needed - probably just as typed converter and not as a top level converter, but didn't bother as we're planning to revamp conversion logic anyway...

PR #470, opened 2025-01-07
fix(tekla): Get Selection Exception

dogukankaratas:
The discussion about this issue can be find in the following tickets:
https://linear.app/speckle/issue/CNX-982/unhandled-exception-while-executing-getselection
https://linear.app/speckle/issue/CNX-743/selection-null-when-its-over-certain-number

This implementation adds a null check for the parts.


oguzhankoral (approved):
Approving with trust

PR #476, opened 2025-01-08
fix(civil): adds top level cogopoint converter

clairekuang:
Cogopoints were not properly supported because the full type name of the class is `Autodesk.**Civil**.DatabaseServices.CogoPoint`, but the inheritance is ` Autodesk.**Autocad**.DatabaseServices.Entity`. 
This means that the top level civil converter is not registering a converter correctly, since cogopoints inherit from autocad entities *not* civil entities and we have no top level entity converter in autocad.

This pr fixes the issue by introducing a top level  `CogoPoint` converter in civil and converting it as a `DataObject` just like all other civil objects. Includes the point name, number, and northing props.

sample model:  https://latest.speckle.systems/projects/2295cb26a0/models/46f44d5e4e

**note** @adamhathcock is making a fix for this on the sdk side to register converters based on the full type name. This was previously resulting in wrong exceptions, eg in this case the Civil Entity converter was successfully registered even though it shouldn't have been, because the inherited base type name (`Entity`) exists in both civil and autocad namespaces

### DIFF
--- a/Converters/Civil3d/Speckle.Converters.Civil3dShared/Civil3dRootToSpeckleConverter.cs
+++ b/Converters/Civil3d/Speckle.Converters.Civil3dShared/Civil3dRootToSpeckleConverter.cs
@@ -34,6 +34,8 @@ public class Civil3dRootToSpeckleConverter : IRootToSpeckleConverter
     object objectToConvert = dbObject;
 
     // check first for civil type objects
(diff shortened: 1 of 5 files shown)


adamhathcock (approved):

adamhathcock:
Related: https://github.com/specklesystems/speckle-sharp-connectors/pull/475

PR #478, opened 2025-01-09
Add custom assembly resolve for Navisworks

adamhathcock:
Add custom assembly resolve for Navisworks to fix Webview conflicts with NW2025


didimitrie (approved):
i trust this works, i don't have navis 2025 installed

PR #481, opened 2025-01-09
fix path encoding

KatKatKateryna:
remove special characters from path after using Uri.AbsolutePath


adamhathcock (changes requested):
I'd want a comment and/or test to say why this is being done

adamhathcock (approved):

PR #482, opened 2025-01-09
fix: stacked wall duplication when sending by view

didimitrie:
See linear ticket - more info and journaling in there.


oguzhankoral (approved):
Even if these conditions has been becoming complex but looks like nothing we can do 🤷

PR #484, opened 2025-01-10
Dimitrie/cnx 916 revit all conversions pass nothing fails 

didimitrie:
A bundled set of fixes, related issues. 

1) Categories in the category filter are now down from ~300 to a more legitimate set of approx 120 that we have high confidence are meaningful - specifically model categories. 

2) We pre-filter elements out that do not make sense to convert anyway, based on the same rules of (1). This means **sending empty views does not succeed anymore**, which is legit. 

3) Pruned views to the ones that actually make sense, ignoring schedules and other view types. 

4) Fixed a performance regression introduced with the data object, whereby we'd be endlessly throwing for all elements with an unsupported location property - which can be numerous, eg. floors

### DIFF
--- a/Connectors/Revit/Speckle.Connectors.RevitShared/HostApp/SupportedCategoriesUtils.cs
+++ b/Connectors/Revit/Speckle.Connectors.RevitShared/HostApp/SupportedCategoriesUtils.cs
@@ -1,0 +1,21 @@
+using Autodesk.Revit.DB;
+
+namespace Speckle.Connectors.Revit.HostApp;
+
+public static class SupportedCategoriesUtils
+{
+  /// <summary>
+  /// Filters out all categories besides Model categories. This utility should be used
+  /// to clean any elements we might want to send pre-conversion as well as in what categories
+  /// to display in our category filter.
+  /// </summary>
+  /// <param name="category"></param>
+  /// <returns></returns>
+  public static bool IsSupportedCategory(Category category)
+  {
+    return (
+        category.CategoryType == CategoryType.Model
+      // || category.CategoryType == CategoryType.AnalyticalModel
+      ) && category.IsVisibleInUI;
+  }
+}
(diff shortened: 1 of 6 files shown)


Connectors/Revit/Speckle.Connectors.RevitShared/HostApp/SupportedCategoriesUtils.cs
```diff
@@ -0,0 +1,21 @@
...
+  /// to clean any elements we might want to send pre-conversion as well as in what categories
+  /// to display in our category filter.
+  /// </summary>
+  /// <param name="category"></param>
+  /// <returns></returns>
+  public static bool IsSupportedCategory(Category category)
+  {
+    return (
+        category.CategoryType == CategoryType.Model
+      // || category.CategoryType == CategoryType.AnalyticalModel
```
didimitrie, line 18: I'm keeping this in here as a reminder/reference

Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/RevitRootObjectBuilder.cs
```diff
@@ -53,15 +53,27 @@ CancellationToken cancellationToken
...
       }
+
+      if (el.Category == null)
+      {
+        continue;
+      }
+
+      if (!SupportedCategoriesUtils.IsSupportedCategory(el.Category))
+      {
+        continue;
```
oguzhankoral, line 68: Would be nice but not sure how much reporting i am talking about
```c#
if (!SupportedCategoriesUtils.IsSupportedCategory(el.Category))
{
  results.Add(new(Status.WARNING, el.UniqueId, el.GetType().Name, null, "Element category is not supported."));
  continue;
}
```

oguzhankoral (changes requested):

Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/RevitRootObjectBuilder.cs
didimitrie: legit, ideally we'd report this once per category - i'll see if i can manage that easily

oguzhankoral (approved):

PR #492, opened 2025-01-13
fix(Navisworks): CNX-895 - Add `ColorProxy` Support for 2D Linework

jsdbroughton:
Currently, all geometry in the Navisworks connector uses RenderMaterialProxies for colour representation. This causes 2D linework to appear uncoloured as it requires ColorProxies instead. This PR adds proper ColorProxy support for 2D elements while maintaining RenderMaterialProxy handling for 3D geometry.

### Changes
- Add new `ColorProxy` handling for 2D linework
- New simplistic `ComApI.InwSimplePrimitivesCB` callback for primitive type detection
- Implement proxy type detection based on geometry type
- Update root object builder to support both proxy types

- Additionally, refactors `BuildAsync` to reduce complexity and clear method steps

## Source

image

## Before

Screenshot 2025-01-13 233327

## After

Screenshot 2025-01-13 233144


oguzhankoral (approved):
Lgtm ❤️

PR #500, opened 2025-01-16
feat(Navisworks): CNX-1064 - Add hierarchy path property to converted Navisworks objects

jsdbroughton:
This adds path information to converted Navisworks objects to provide better context about their location in the model hierarchy. Instead of returning a hierarchical data structure, each object in the list is flat but has hierarchical information in the `path`.

Closes CNX-1064 & CNX-1067

### Key changes:

- Added a `path` property to all converted objects, showing their hierarchical location in the model
- Created `DisplayPathHelper` to generate standardized paths by traversing meaningful ancestors
- Modified `NavisworksRootObjectBuilder` to handle path generation for both individual and grouped geometry objects
- Added logic to find meaningful ancestor names for grouped geometry nodes
- Ensured paths don't duplicate the object's name, which is already shown in the name property 
- Paths follow the format: "file.nwd > Assembly > Subassembly"

#### Example:
```
name: "Hood1A-1"
path: "snowmobile.nwd > Hood-2" 
```

This change improves object identification and filtering capabilities by providing a clear hierarchical context for each object in the converted model.

Manual Tests:
- Source object in the selection tree
   image

- As name and path as published
   image


bimgeek (approved):
i can confirm path is added to objects.
image

PR #513, opened 2025-01-22
Use Event Aggregator with Revit

adamhathcock:
replaces https://github.com/specklesystems/speckle-sharp-connectors/pull/502


Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
```diff
@@ -305,9 +307,9 @@ private bool HaveUnitsChanged(Document doc)
     return false;
   }
 
-  private async Task PostSetObjectIds()
+  private async Task PostSetObjectIds(object _)
```
oguzhankoral, line 310: a bit confused why we have this object argument

oguzhankoral: got it!

JR-Morgan (approved):

oguzhankoral (approved):

PR #523, opened 2025-01-27
Use Event Aggregator with Navisworks

adamhathcock:


jsdbroughton:
Looks legit but I'll test runtime. 

Not introduced by this PR but thinking harder about something else about repeated one time binding on send. This is what I was going to test again as mentioned: https://discord.com/channels/726756379083145269/751053204103495711/1332369121203196014

```csharp
// In NavisworksDocumentEvents.cs
_eventAggregator.GetEvent<IdleEvent>().OneTimeSubscribe(
    nameof(NavisworksDocumentEvents), 
    ProcessModelStateChangeAsync
);
```

Might want to verify the cleanup behavior with one-time subscriptions if state changes during processing.

Maybe nothing. And not new.

jsdbroughton (approved):
Successfully tested for non-regressions (including the comment mentioning other issue)

PR #531, opened 2025-01-28
Refactor Converters to have feedback for Success, Warnings, Errors without using exceptions

adamhathcock:
Starting with Autocad


JR-Morgan (changes requested):
see https://discord.com/channels/726756379083145269/751053204103495711/1334105597632581673

PR #538, opened 2025-01-30
fix(etabs): catch tendon and link objects as warning

bjoernsteinhagen:
## Description

- Link to [CNX-1138](https://linear.app/speckle/issue/CNX-1138/unexpected-object-selection)
- `ETABS` as a building module should not have `REBAR`, `LINK` or `TENDON` objects modelled (i.e. we won't support this workflow). However, for the event that these objects are present in the model, we note the proper conversion success and don't throw a conversion error in the UI.

## User Value
User is informed of which objects aren't converted and why. Not that the whole conversion fails.

## Changes:

- `GetSpeckleApplicationId` for all possible types
- `NotImplementedException` for `LINK` and `TENDON`
- Marked as `WARNING` in the conversion result

### Before

image

### After

image

## Checklist:


dogukankaratas (approved):
seems good to me

PR #548, opened 2025-02-03
fix: prevents fatal crash on old model cards

didimitrie:
Somehow, a failing `.NotNull()` in `RunExpirationChecks` creates a fatal crash down the line. With this fix in, I can't reproduce the linear issue attached. 

https://github.com/user-attachments/assets/0c10d175-1464-4f00-8cd5-c3cee502c674

**Before merging (or after) I would really want @bimgeek to test this, as I could not reproduce the exact issue you had!**


adamhathcock (approved):

bimgeek (approved):
doesn't crash anymore!

PR #556, opened 2025-02-04
feat(autocad/rhino): add name and properties

clairekuang:
## Description

Adds `name` and `userstrings` to Rhino via a `PropertiesExtractor` and `UserStringsExtractor` class in the connector (because converter has no RootToSpeckle class)
Adds `extensiondictionary` to Autocad via a `PropertiesExtractor` and `ExtensionDictionaryExtractor` class in the converter

Also updates object attributes event listener to include additional name and user string props in rhino

## User Value

Enables data extraction workflows for rhino and autocad

## Changes:

- Rhino connector
- Autocad Converter
- Civil3d converter

## Screenshots:

Rhino model: https://latest.speckle.systems/projects/2295cb26a0/models/63ea5caf19
Autocad model:

## Validation of changes:

- tested in rhino the name and user string props of point, curve, instance, brep, and subd objects


adamhathcock (commented):
I'd like [GenerateAutoInterface] and interfaces used.

Ideally there would be tests for this extra logic against an object

Connectors/Rhino/Speckle.Connectors.RhinoShared/HostApp/Properties/PropertiesExtractor.cs
```diff
@@ -0,0 +1,36 @@
...
+
+    return properties;
+  }
+
+  private void AddDictionaryToPropertyDictionary(
+    Dictionary<string, object?>? entryDictionary,
+    string entryName,
+    Dictionary<string, object?> propertyDictionary
+  )
+  {
```
oguzhankoral, line 30: i do not believe we need this function at all, i would just put if logic to proper place. Also, It says `Add...To...` but it doesnt guarntee it will add since it has if statement. At least i would rename if to `Try...Add...To`

Connectors/Rhino/Speckle.Connectors.RhinoShared/HostApp/Properties/UserStringsExtractor.cs
```diff
@@ -0,0 +1,34 @@
...
+using Rhino.DocObjects;
+
+namespace Speckle.Connectors.Rhino.HostApp.Properties;
+
+/// <summary>
+/// Extracts user strings out from a rhinoobject. Expects to be scoped per operation.
+/// </summary>
+public class UserStringsExtractor
+{
+  public UserStringsExtractor() { }
```
oguzhankoral, line 11: not sure it is necessary

Converters/Autocad/Speckle.Converters.AutocadShared/ToSpeckle/Properties/PropertiesExtractor.cs
```diff
@@ -0,0 +1,38 @@
...
+    return properties;
+  }
+
+  private void AddDictionaryToPropertyDictionary(
+    Dictionary<string, object?>? entryDictionary,
+    string entryName,
+    Dictionary<string, object?> propertyDictionary
+  )
+  {
+    if (entryDictionary is not null && entryDictionary.Count > 0)
```
oguzhankoral, line 33: same here

oguzhankoral (changes requested):
I have general comment that we can simplify some logics, null checks can be extracted out from functions since we check them first and return null directly, i would do this checks at the first place, it a bit isolates the reasoining to return null

Converters/Autocad/Speckle.Converters.AutocadShared/ToSpeckle/Properties/PropertiesExtractor.cs
didimitrie: also fixed

didimitrie (approved):

adamhathcock (approved):

clairekuang:
@oguzhankoral retested after threading pr merge, this pr is still looking good

oguzhankoral (approved):
tested and looks good

PR #558, opened 2025-02-04
feat(revit): add density to material quantities

bjoernsteinhagen:
## Description

- Related to [CNX-1088](https://linear.app/speckle/issue/CNX-1088/revit-add-structural-and-thermal-properties-to-materials).
- Stems from user request to extend parameters on material quantities: [Accessing Material Density Parameter Value](https://speckle.community/t/accessing-material-density-parameter-value/16026).

## User Value

Material Quantities can be used to calculate object mass; however, in order to compute mass, users need to have `density` made available to them.

## Changes:

- Implemented `StructuralMaterialAssetExtractor.cs` to handle extraction of `density` from `StructuralAsset`
- Since `area` and `volume` are scaled from internal units to model units, the same is done for `density` to avoid confusion
- Display units added to the properties for clarity
- `density` per `StructuralAsset` cached locally in the `MaterialQuantitiesToSpeckle.cs`
- Since the current scope was limited to `density` ONLY, no need to create `MaterialProxy` as initially thought. Should the extracted parameters grow, this must be re-evaluated / refactored!
- Note if anything on the extracting parameters from structural asset fails, the `density` will simply not be appended to the object. This isn't seen as a failure case.

### Before:

image

### After:

image

## Validation of changes:

Tested on various structural and architectural models for Revit versions 2022-2024:

- [Speckle Model - Revit 2022](https://latest.speckle.systems/projects/99d30f721c/models/d66543fbb4)
- [Speckle Model - Revit 2023](https://latest.speckle.systems/projects/99d30f721c/models/6c9fd3450b)
- [Speckle Model - Revit 2024](https://latest.speckle.systems/projects/99d30f721c/models/9cd7d82662)

## Checklist:


KatKatKateryna (approved):
Looks good to me, including a nice fix for the units for Area and Volume.
From my limited knowledge of Revit materials the next comment might not make sence so correct me if I'm wrong: 
Would it make sense to check "if (_converterSettings.Current.Document.GetElement(matId) is DB.Material material)" even before assigning Area and Volume? It seems for that step ".GetElement(matId)" is already expected to be a Material

bjoernsteinhagen:
> Looks good to me, including a nice fix for the units for Area and Volume. From my limited knowledge of Revit materials the next comment might not make sence so correct me if I'm wrong: Would it make sense to check "if (_converterSettings.Current.Document.GetElement(matId) is DB.Material material)" even before assigning Area and Volume? It seems for that step ".GetElement(matId)" is already expected to be a Material

That's a good question. This was taken over from earlier definition. From what I know, the `GetElement` is not type specific and to access `MaterialCategory` and `MaterialClass`, we need to type cast to `DB.Material` and this is the safest way. But I'm not hundred percent sure on this.

PR #564, opened 2025-02-05
fix(rhino): bakes render materials on receive instead of materials

clairekuang:
## Description

Previously, we were baking Rhino Materials (old class) on receive instead of Rhino Render Materials (new class). This means that we were creating *a material for every object with a material assigned* instead of * a single material per material proxy*. This is because the Material Table in Rhino creates a material per object, resulting in diff ids when sent again, even though this is not apparent in the UI.

By creating Render Materials, we are only making one render material per proxy.

## User Value

Prevents bad send data with duplicatd material proxies, and speeds up sending received models with materials.

## Changes:

- rhino material baker
- rhino material unpacker
- pre receive deep clean to purge previously received materials

## Screenshots:

Revit model: 44 render material proxies
{498B55EE-5CA1-4207-9623-81A0C5AA3549}

Before: 5700 render material proxies
https://latest.speckle.systems/projects/3f895e614f/models/43cb043ee7
{9C1CBE67-D338-49EA-92C7-9CD808B5E85A}

After: 34 render material proxies (some objects failed on receive)
https://latest.speckle.systems/projects/3f895e614f/models/43cb043ee7
{0FA5E780-D5B8-4850-B2A4-0F2D36F28632}

## Validation of changes:

- tested by receiving and then sending both rhino and revit commits
- checked that all relevant render materials are purged before receive


oguzhankoral (approved):
LGTM! and i am happy we got rid of material + render material logic

PR #517, opened 2025-01-23
Use Event Aggregator with Autocad/Civil

adamhathcock:

PR #570, opened 2025-02-06
Removes EA from Autocad

adamhathcock:
opposite of https://github.com/specklesystems/speckle-sharp-connectors/pull/517


oguzhankoral:

PR #571, opened 2025-02-06
Removes EA from Navisworks

adamhathcock:
Opposite of https://github.com/specklesystems/speckle-sharp-connectors/pull/523


oguzhankoral:

PR #417, opened 2024-11-28
Introduce Event aggregator

adamhathcock:
https://www.notion.so/speckle/Event-Aggregation-174b78fc7aa6806a86e7f451fa7c4b96

Decouples exception handler from browser bridge.
IdleManager is now a OneTime event with IDs.
Rhino is converted.
Global Events (Doc changed) are converted.

More events to come!

PR #573, opened 2025-02-06
Removes EA from Tekla, a bit of base work needed for doc store removal

adamhathcock:
removes EA from Tekla and a bit of base work for Document Store

undoes some of https://github.com/specklesystems/speckle-sharp-connectors/pull/417


oguzhankoral:

PR #581, opened 2025-02-11
refactor(revit): properties cleanup

clairekuang:
## Description

Makes all revit element properties consistent on send:
- removes all properties dynamically attached to converted objects that were a *duplicate* of a parameter
- moves all dynamically attached element-level properties like `workset`, `elementid`, `level`, and `builtincategory` to the top level of the `properties` dictionary.
- removes the dynamically attached `curve` property of grids into the grid display value (resulting in minor refactor of display value extractor, and adding grids back into supported categories)

Also fixes a bug where we were still sending an empty commit if only unsupported category objects are selected.

## User Value

makes consuming properties on all revit objects (data objects + geometry) consistent

## Changes:

- revit converter

## Screenshots:

Before:
Sample house model: https://latest.speckle.systems/projects/2295cb26a0/models/8358cdcd15@313bd10d67
image
image

After:
Sample house model: https://latest.speckle.systems/projects/2295cb26a0/models/8358cdcd15
Grids added back in with properties: https://latest.speckle.systems/projects/2295cb26a0/models/83088cb067
image
image
{607DD5CE-A37D-406B-B4F3-F5047E351075}

## Validation of changes:

- sent sample house and inspected props 

## Checklist:

- [ ] I have updated or added relevant documentation in figjam board


Connectors/Revit/Speckle.Connectors.RevitShared/HostApp/SupportedCategoriesUtils.cs
```diff
@@ -5,26 +5,40 @@ namespace Speckle.Connectors.Revit.HostApp;
...
+    if (category is null || !category.IsVisibleInUI)
+    {
+      return false;
+    }
+
+    switch (category.CategoryType)
+    {
+      case CategoryType.Annotation:
+        return
+#if REVIT_2023_OR_GREATER
```
didimitrie, line 25: note to double check on gridlines

didimitrie: (claire says dis not working as expected)

didimitrie (approved):

didimitrie:
finally tested! i think we need to cuddle up to @jsdbroughton more for automate purposes on this.

PR #583, opened 2025-02-12
Dimitrie/cnx 1238 rhino 8 restarting during an operation throws with key not

didimitrie:
Fixes a rhino issue that was introduced by https://github.com/specklesystems/speckle-sharp-connectors/pull/557

@adamhathcock this fix is probably not cool, but happy if you take over with a more proper one if it's really needed to have disposal on `ICancellationItem`s


didimitrie:
follow up: my tldr would be if that there's no real/reported/clear issue coming from NOT disposing our tokens, i'm happy for the old way to be back

adamhathcock (changes requested):
I think this will probably be more fixed by an SDK PR:https://github.com/specklesystems/speckle-sharp-sdk/pull/225

I think it's just not downloading everything expected on a cancel.  Kat was seeing this too.

Not disposing seems like a bad idea.

didimitrie:
closing, resolved by @adamhathcock 's pr :)

PR #587, opened 2025-02-13
fix: (cancellation manager should return cancelled token instead of throwing exception)

adamhathcock:
## Description

CancellationManager shouldn't ever throw.  This results in error behavior we don't want.  If a card's cancellation is missing, assume it should be cancelled.

## Validation of changes:

can follow repo steps with no errors: https://linear.app/speckle/issue/CNX-1238/rhino-8-restarting-during-an-operation-throws-with-key-not-present-in

### DIFF
--- a/Sdk/Speckle.Connectors.Common/Cancellation/CancellationManager.cs
+++ b/Sdk/Speckle.Connectors.Common/Cancellation/CancellationManager.cs
@@ -27,18 +27,9 @@ public class CancellationManager : ICancellationManager
 
   public int NumberOfOperations => _operationsInProgress.Count;
 
-  /// <summary>
-  /// Get token with registered id.
-  /// </summary>
-  /// <param name="id"> Id of the operation.</param>
-  /// <returns> CancellationToken that belongs to operation.</returns>
-  public CancellationToken GetToken(string id) => _operationsInProgress[id].Token;
+  //if we can't find it then it must be cancelled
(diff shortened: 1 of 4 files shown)


Sdk/Speckle.Connectors.Common/Cancellation/CancellationManager.cs
```diff
@@ -27,18 +27,10 @@ private sealed class CancellationItem(CancellationManager manager, string id) :
...
   public int NumberOfOperations => _operationsInProgress.Count;
 
-  /// <summary>
-  /// Get token with registered id.
-  /// </summary>
-  /// <param name="id"> Id of the operation.</param>
-  /// <returns> CancellationToken that belongs to operation.</returns>
-  public CancellationToken GetToken(string id) => _operationsInProgress[id].Token;
+  //if we can't find it then it must be cancelled
+  private CancellationToken GetToken(string id) =>
```
oguzhankoral, line 31: why otherwise it generates new token, how it relates to `id` that passed by function argument

adamhathcock: it's a comment: `//if we can't find it then it must be cancelled`

Doing `new (true)` on CancellationToken struct will mean it's always cancelled.  It could only be missing if previously disposed.  I don't know how the cards are passing an ID that already cancelled but they are.

didimitrie: ah, so for later requests we're passing an already cancelled token, which should ensure right behaviour. checking now

PR #603, opened 2025-02-19
fix: some Revit perf

adamhathcock:
Part of https://linear.app/speckle/issue/CNX-1324/improving-send-and-receive-perf

nothing earth shattering but identified some hotspots to increase some speed.  Benchmarks didn't increase a lot as most time is spent in Revit API
image


Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/MeshConverterToHost.cs
```diff
@@ -63,12 +55,11 @@ out var mappedElementId
       {
         // Non-planar quads will be triangulated as it's more desirable than `TessellatedShapeBuilder.Build`'s attempt to make them planar.
         // TODO consider triangulating all n > 3 polygons that are non-planar
-        var triPoints = new List<XYZ> { points[0], points[1], points[3] };
```
adamhathcock, line 66: need to fix this

Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/MeshConverterToHost.cs
```diff
@@ -125,14 +123,14 @@ private XYZ[] ArrayToPoints(IList<double> arr, string units)
     }
 
     XYZ[] points = new XYZ[arr.Count / 3];
-    var fTypeId = _scalingServiceToHost.UnitsToNative(units) ?? UnitTypeId.Meters;
+    var fTypeId = scalingServiceToHost.UnitsToNative(units);
```
oguzhankoral, line 126: why remove the fallback here, I would not touch it

oguzhankoral (changes requested):

Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/MeshConverterToHost.cs
adamhathcock: It never returns null so it's redundant.  `UnitsToNative` throws an exception if the value is bad.

adamhathcock:
This doesn't work as large amounts of meshes will overflow
Screenshot 2025-03-06 164222

PR #608, opened 2025-02-19
fix(CI): No files, no success

JR-Morgan:
Make the CI fail the jobs if it fails to publish artefacts or installers


oguzhankoral (commented):
I believe this will cause failing on release branch if installers are failed?

JR-Morgan:
@oguzhankoral correct, if the `connector-installers` workflow fails, then so will the `deploy-installers` job on this repo.
The artifacts are set to stick around for 1 day. In error case.

oguzhankoral (approved):
❤️

PR #609, opened 2025-02-19
fix: Autocad perf - Adds early exit to idle manager to shortcut call check

adamhathcock:
Part of https://linear.app/speckle/issue/CNX-1324/improving-send-and-receive-perf

Since Autocad calls object changed on each receive, this shaves off 7 seconds off my test.  Also fixes a namespace

Should talk about Transactions  vs StartStop ones

before:
Screenshot 2025-02-19 161647

after:
Screenshot 2025-02-19 173121


oguzhankoral:
Since this PR touches the IdleManager which would effect the all connectors that means need to be tested widely. I would rather to have this imporevements in after end of Feb release. Thats why i marked it as draft.

oguzhankoral (approved):
Tested and looks good

PR #612, opened 2025-02-20
feat(tekla): speckle ribbon

dogukankaratas:
adds a speckle ribbon to tekla.

currently, we are launching our connector from Applications & Components tab

image

with this pr, this option will be still available, and also adds the ribbon for easy usage.

image

the reason that I'm using the old logo is, the new logo cannot be rendered in tekla. it should be a svg format and svg formats with patterns creates a black shadow.

renderissue

PS: I will also update the installer accordingly.


oguzhankoral (changes requested):
@bimgeek could you provide icon to Dogukan as tekla requests? Don't we have vectoral DUI3 logo? if not, I wouldnt bother

bimgeek:
> @bimgeek could you provide icon to Dogukan as tekla requests? Don't we have vectoral DUI3 logo? if not, I wouldnt bother

@oguzhankoral I already did.

bimgeek:
@oguzhankoral here's the svg file:

tekla-icon@32

Let me know if it does the job. I converted png to SVG so it's a bit dirty.

dogukankaratas:
Seems good. Thanks @bimgeek !

image

bjoernsteinhagen (approved):

PR #614, opened 2025-02-21
Arcgis detailed progress and cancellation

KatKatKateryna:
## Description
By now, layer conversion was treated as a single operation, so the progress was counting an entire layer as 1 object. This prevented a detailed progress update, and cancellation of the operation didn't actually take effect until the entire layer was converted. This is particularly a problem for large layers (with hundreds of thousands of features).

## User Value
- User can see the progress feature by feature
- User can cancel operation at any stage of layer conversion
image
- Also works with Pointclouds: 
image

## Changes:
1. Count total features per layer, use the sum for progress bar.
2. Insert cancellation token check, current feature count and progress update inside the layer conversion, so it's updates before every feature (instead of before every layer)

## Checklist:


adamhathcock (approved):
Looks good to have more fine-grained progress

PR #621, opened 2025-02-25
enhancement(Navisworks): CNX-1068 - Reduce Transmittal of internal navisworks properties and model properties.

jsdbroughton:
- Excluded "Transform" category from processing.
- Added a list of excluded properties: "Hidden", "Required", "Internal_Type".
- Improved handling for the "Item" category to add non-excluded properties directly to root level.
- Ensured only non-empty property dictionaries are processed.

### DIFF
--- a/Converters/Navisworks/Speckle.Converters.NavisworksShared/DataHandlers/BasePropertyHandler.cs
+++ b/Converters/Navisworks/Speckle.Converters.NavisworksShared/DataHandlers/BasePropertyHandler.cs
@@ -17,17 +17,35 @@ public abstract class BasePropertyHandler(
 
     if (propertySets != null)
     {
-      foreach (var category in propertySets.Where(c => c.Key != "Internal"))
+      foreach (var category in propertySets.Where(c => c.Key != "Internal" && c.Key != "Transform"))
       {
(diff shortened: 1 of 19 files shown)


Converters/Navisworks/Speckle.Converters.NavisworksShared/DataHandlers/BasePropertyHandler.cs
```diff
@@ -17,17 +17,35 @@ ModelPropertiesExtractor modelPropertiesExtractor
 
     if (propertySets != null)
     {
-      foreach (var category in propertySets.Where(c => c.Key != "Internal"))
+      foreach (var category in propertySets.Where(c => c.Key != "Internal" && c.Key != "Transform"))
       {
+        var excludedProperties = new List<string> { "Hidden", "Required", "Internal_Type" };
```
oguzhankoral, line 22: I would put this into outside of for each loop since we do not populate it, there is no need to generate the list every time

jsdbroughton: Changed to a read-only class property

oguzhankoral (approved):

PR #622, opened 2025-02-25
enhancement(Navisworks): CNX-972 - less aggressive object merging

jsdbroughton:
## Description

Previously, our GeometryNodeMerger grouped only anonymous geometry nodes with the same parent path. This PR refactors that logic to consider material properties and generate a unique signature for merging. This future-proofs the connector by adding additional properties to the signature generation process without reworking the entire grouping logic.

### Key updates include:

1. Enhanced Grouping Logic
   - `GroupSiblingGeometryNodes` now appends a material-based signature to that path instead of grouping nodes solely by the parent path.
   - Nodes with the same parent path and identical material signatures will be merged, reducing redundant geometry.

2. Signature Generation

   - New private methods:
      - `GenerateSignature(NAV.ModelItem node)`: aggregates relevant material properties (colour, transparency, etc.).
      - `AddMaterialProperties(...)`: a flexible entry point to include more properties in the future (e.g., textures or custom user data).
      - `GetSignature(...)`: constructs a stable string from sorted property key-value pairs, then hashes it.
      - `GetHash(...)`: produces a short, non-cryptographic hash for minimal collisions.

3. Material Name Extraction

   - `GetMaterialName(NAV.ModelItem node)` checks the “`Item`” and “`Material`” property categories to retrieve a name if available.
 
This refactor increases maintainability by centralizing property-based merging, making expanding or modifying what defines “identical” geometry easier.

## User Value

- Optimized Model Sends: Merging geometry nodes with identical material properties reduces data duplication and sending time. However the prior implementation lost material definitions between sibling geometries.
- Easier Maintenance: The refactor includes a straightforward, extensible signature generation approach, ideal for future additions (e.g., extra metadata or layering info).
- Reduced Code Duplication: A single entry point to gather and process properties for grouping.

## Changes:

1. `GeometryNodeMerger.cs`

- Replaced the old single-line `GroupSiblingGeometryNodes` method with a more robust approach that:
- Collects each node’s parent path.
- Appends a hashed signature of relevant material properties to form a composite key.
- Groups nodes accordingly in a dictionary, ensuring we only merge sets with identical signatures.
- Added helper methods (`GenerateSignature`, `AddMaterialProperties`, `GetSignature`, `GetHash`, and `GetMaterialName`).

### Source
image

This shows that all the geometry nodes are sibling nodes.

### Before
image

Arbitrarily the nodes were merged into two grouped node of sibling objects (lines and meshes)

### After
Screenshot 2025-02-25 215558

The current implementation differentiates siblings by material appearance.

image
Diff shows the impact and additional unique objects for the same source.

image
`applicationId` modified from the basic path description pattern 0/1/23.... to suffix `::signature_hash` where merged nodes exist.

## Validation of changes:

Manual testing across multiple formats and model hierarchies.

## Checklist:

### DIFF
--- a/Connectors/Navisworks/Speckle.Connectors.NavisworksShared/Operations/Send/GeometryNodeMerger.cs
+++ b/Connectors/Navisworks/Speckle.Connectors.NavisworksShared/Operations/Send/GeometryNodeMerger.cs
@@ -1,4 +1,5 @@
 ﻿using Speckle.Connector.Navisworks.Services;
+using Speckle.Converter.Navisworks.Constants;
 
 namespace Speckle.Connector.Navisworks.Operations.Send;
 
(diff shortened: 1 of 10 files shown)


Connectors/Navisworks/Speckle.Connectors.NavisworksShared/Operations/Send/GeometryNodeMerger.cs
```diff
@@ -7,16 +8,156 @@ namespace Speckle.Connector.Navisworks.Operations.Send;
...
+    return hashInput.Length == 0 ? "empty" : GetHash(hashInput.ToString(), hashLength);
+  }
+
+  /// <summary>
+  /// Generates a non-cryptographic hash from the input string. Mostly to avoid MD5.Create() warnings.
+  /// </summary>
+  /// <param name="input">The input string to hash.</param>
+  /// <param name="length">The length of the returned hash string (default: 8 characters).</param>
+  /// <returns>A non-cryptographic hash string of the specified length.</returns>
+  private static string GetHash(string input, int length = 8) =>
```
adamhathcock, line 130: Don't do this...use MD5 if you mean to or a SHA version

adamhathcock (commented):
This looks like it will be very slow because it's generating and appending strings a lot.  I'd like to understand the point of this.

Connectors/Navisworks/Speckle.Connectors.NavisworksShared/Operations/Send/GeometryNodeMerger.cs
jsdbroughton: Fair. I started with MD5.Create() and got security flagged. It's not needed to be tremendously secure 🫠

adamhathcock: Can just pragma around it if it doesn't need to be secure and MD5 is good enough

jsdbroughton:
> This looks like it will be very slow because it's generating and appending strings a lot.  I'd like to understand the point of this.

I tried to be descriptive of the why in the PR. But here goes...

On ingestion from certain applications (not Revit) there is a preponderance sometimes of many many anonymous geometry nodes. It doesn't make a great deal of UX sense to have all of these nodes available in the tree without any particular metadata to them. The containing parent (sometimes grandparent) node is usually where important data is so merging geometry is the strategy. I coalesce all of those geometry nodes into a single display value per parent. That's already in place. What this PR does is to be a bit smarter because it was a merging all geometry nodes and disregarding differences in material--then thinking as an extension of that there may be situations where we should be less aggressive on other terms-- so this was a signature generator to group by. Completely happy for someone to think of a better solution.

adamhathcock:
I think I need a more hand holding meeting to show me things.  I'll let @oguzhankoral or @clairekuang decide though

oguzhankoral:
Idea is legit. I do something similar in Sketchup to group meshes by material and properties under components to decrease footprint and select them together. (6 faces turn into 3 meshes).

image

Codewise, if there is no any performace/behavior penalty i am happy to merge it away. You tell me @jsdbroughton

PR #623, opened 2025-02-26
fix(Revit) improving revit selection perf

adamhathcock:
Details on Linear issue: https://linear.app/speckle/issue/CNX-1323/improving-revit-selection-perf

Increase perf by reducing allocations for large selections


oguzhankoral (approved):
Tested and LGTM even if we are having our own `ConcurrentHashSet`

PR #624, opened 2025-02-26
fix: Use a concurrent dictionary to cache types since the converter is effectively a singleton

adamhathcock:
## Description

DiscriminatedObjectConverter is effectively a singleton.  To cache types, use a concurrent dictionary rather than changing the lifetime.

## User Value

stops random errors


oguzhankoral (approved):
Tested and looks good!

PR #642, opened 2025-03-04
Test release for IFC Importer

JR-Morgan:


adamhathcock (approved):
Not a public release, needed for IFC updating

JR-Morgan:
Note: this did not trigger a release, because of the way nuget only increments the version when a real release is made. currently we have no nice way to trigger nugets independently from connectors 😢

PR #643, opened 2025-03-05
Rhino region conversions

KatKatKateryna:
## Description
Conversion toSpeckle and toHost for Hatches in Rhino. 

New Region geometry class in the hostApps is usually called Region or Polygon (pure geometry), or Hatch ("region" with a pattern or gradient fill). The property "hasHatchPattern" on Region objects indicates which one it is. If the host app only supports one type, conversion defaults to that 1 type. Like in case with Rhino: it only supports Hatch, so any Region from Speckle is received as Hatch in Rhino.

Currently the displayValue is a list of combined boundary and loop curves. We hope that in the future the Viewer can support this geometry and create a fill in the Viewer, without the need to generate and send an extra Mesh. 

Sent:
image

In Speckle https://app.speckle.systems/projects/16ce7b208c/models/8b96943c98@86a3023149 :
image

Received back in Rhino (hatch styles ignored):
image

## Checklist:


Converters/Rhino/Speckle.Converters.RhinoShared/ToSpeckle/TopLevel/HatchObjectToSpeckleTopLevelConverter.cs
```diff
@@ -0,0 +1,81 @@
...
+
+    SOG.Box bbox = GetBboxFromHatch(hatch);
+
+    return new SOG.Region
+    {
+      boundary = boundary,
+      innerLoops = innerLoops,
+      hasHatchPattern = true,
+      bbox = bbox,
+      displayValue = allCurves.Cast<Base>().ToList(),
```
clairekuang, line 54: should set this displayvalue to null instead, no need to duplicate curves

clairekuang (changes requested):

Converters/Rhino/Speckle.Converters.RhinoShared/ToSpeckle/TopLevel/HatchObjectToSpeckleTopLevelConverter.cs
```diff
@@ -0,0 +1,81 @@
+﻿using Rhino.DocObjects;
+using Speckle.Converters.Common;
+using Speckle.Converters.Common.Objects;
+using Speckle.Objects;
+using Speckle.Sdk.Models;
+
+namespace Speckle.Converters.Rhino.ToSpeckle.TopLevel;
+
+[NameAndRankValue(typeof(HatchObject), NameAndRankValueAttribute.SPECKLE_DEFAULT_RANK)]
+public class HatchObjectToSpeckleTopLevelConverter : IToSpeckleTopLevelConverter
```
clairekuang, line 10: use the RhinoObjectToSpeckleTopLevelConverter pattern here, since `Hatch` is a `HatchObject` and the RG.Hatch geometry can be accessed via `hatchobject.Geometry`.

This means this conversion logic should be in the raw converter. this will be important when we need to inject the raw converters into eg grasshopper.

Converters/Rhino/Speckle.Converters.RhinoShared/ToSpeckle/TopLevel/HatchObjectToSpeckleTopLevelConverter.cs
```diff
@@ -0,0 +1,81 @@
...
+    _curveConverter = curveConverter;
+  }
+
+  public Base Convert(object target)
+  {
+    var hatchObject = (HatchObject)target;
+    var hatch = (RG.Hatch)hatchObject.Geometry;
+
+    // get boundary and inner curves
+    RG.Curve rhinoBoundary = (hatch).Get3dCurves(true)[0];
```
clairekuang, line 36: the return type and method documentation implies there may be multiple boundary curves. could you validate if this is true or not?

Converters/Rhino/Speckle.Converters.RhinoShared/ToSpeckle/TopLevel/HatchObjectToSpeckleTopLevelConverter.cs
```diff
@@ -0,0 +1,81 @@
...
+      boundary = boundary,
+      innerLoops = innerLoops,
+      hasHatchPattern = true,
+      bbox = bbox,
+      displayValue = allCurves.Cast<Base>().ToList(),
+      units = _settingsStore.Current.SpeckleUnits,
+    };
+  }
+
+  private SOG.Box GetBboxFromHatch(RG.Hatch hatch)
```
clairekuang, line 59: tbh I'd delete this- we aren't using bbox for anything, this is wasted conversion so you can just pass a null

clairekuang (changes requested):

Converters/Rhino/Speckle.Converters.RhinoShared/ToSpeckle/TopLevel/HatchObjectToSpeckleTopLevelConverter.cs
KatKatKateryna: I see, I was looking at the Brep TopLevel conversion which was fully done there. But now I can see other geometries are following the pattern 👍

Converters/Rhino/Speckle.Converters.RhinoShared/ToSpeckle/TopLevel/HatchObjectToSpeckleTopLevelConverter.cs
KatKatKateryna: It uses the same .Get3dCurves method for outer and inner curves (you specify the difference with bool input variable), so the return type is enumerable. But outer boundary can only be 1

clairekuang (changes requested):
Don't forget to update sdk nuget to include the region class, so build will pass.
Also, add some screenshots for the received hatches! Probably would be good to update the objects file in rhino with different types of hatches for testing

KatKatKateryna:
> Don't forget to update sdk nuget to include the region class, so build will pass. Also, add some screenshots for the received hatches! Probably would be good to update the objects file in rhino with different types of hatches for testing

Added a new ticket to update Objects model for all apps sending new Region class:
https://linear.app/speckle/issue/CNX-1405/update-sample-object-models-from-supported-send-connectors-to-include

KatKatKateryna:
Objects model updated: https://latest.speckle.systems/projects/2295cb26a0/models/bd1fd98086@fda1477843
image

Converters/Rhino/Speckle.Converters.RhinoShared/ToHost/Raw/RegionToHostConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+  {
+    List<RG.Curve> rhinoCurves = new() { _curveConverter.Convert(target.boundary) };
+    foreach (var loop in target.innerLoops)
+    {
+      // TODO: add extra checks depending on edge case Regions sent from other apps (e.g. not closed)
+      rhinoCurves.Add(_curveConverter.Convert(loop));
+    }
+    var tolerance = _settingsStore.Current.Document.ModelAbsoluteTolerance;
+    // .Create method returns array, but in case of Speckle Region we always expect only 1 converted Hatch
+    RG.Hatch result = RG.Hatch.Create(rhinoCurves, 0, 0, 1, tolerance)[0];
```
clairekuang, line 37: This doesn't look quite right. Use the method I linked instead, so you can set the outer and inner loops directly. The method you use can return an array of hatches, especially if the curve is self-intersecting. Since we do not prevent people from sending "bad" hatches, you shouldn't assume there is only one returned https://developer.rhino3d.com/api/rhinocommon/rhino.geometry.hatch/create#(plane,curve,ienumerable%3Ccurve%3E,int,double,double)

Converters/Rhino/Speckle.Converters.RhinoShared/ToHost/Raw/RegionToHostConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+  /// </summary>
+  /// <param name="target">The Speckle Region geometry to convert.</param>
+  /// <returns>The converted Rhino Hatch.</returns>
+  /// <remarks>⚠️ This conversion does NOT perform scaling.</remarks>
+  public RG.Hatch Convert(SOG.Region target)
+  {
+    List<RG.Curve> rhinoCurves = new() { _curveConverter.Convert(target.boundary) };
+    foreach (var loop in target.innerLoops)
+    {
+      // TODO: add extra checks depending on edge case Regions sent from other apps (e.g. not closed)
```
clairekuang, line 32: remove this todo if this isn't done in the pr: I wouldn't worry about object validation atm, if there's a bad hatch then let it throw on create below

Converters/Rhino/Speckle.Converters.RhinoShared/ToSpeckle/Raw/HatchToSpeckleConverter.cs
```diff
@@ -0,0 +1,47 @@
...
+  public SOG.Region Convert(RG.Hatch target)
+  {
+    // get boundary and inner curves
+    RG.Curve rhinoBoundary = target.Get3dCurves(true)[0];
+    RG.Curve[] rhinoLoops = target.Get3dCurves(false);
+
+    ICurve boundary = _curveConverter.Convert(rhinoBoundary);
+    List<ICurve> innerLoops = rhinoLoops.Select(x => _curveConverter.Convert(x)).ToList();
+
+    List<ICurve> allCurves = new List<ICurve> { boundary }
```
clairekuang, line 35: this is a bit inefficient, since you are creating a new list with `concat` and then calling `tolist()` again. You can just use an `addrange` to modify your allcurves list.

Actually, you aren't using the allcurves list at all. Just delete lines 35-37

clairekuang (changes requested):

Converters/Rhino/Speckle.Converters.RhinoShared/ToHost/Raw/RegionToHostConverter.cs
KatKatKateryna: This method will require a Hatch Plane as an input, which we were not storing at any point (plane was stored as a part of Bbox, but was deleted at request in one of the previous commits). If Plane is something necessary, we can revise Region class to include Plane in the constructor, or simply return the Bbox conversion (possibly non-nullable), and we will get the plane as a part of it. We can quickly sync to decide what's best 🙌 
The simplest option if to keep the method which is used now, possibly adding a check that exactly one Hatch was generated, 0 or >1 means an error:
image

PR #647, opened 2025-03-05
Nuget publish only on release

JR-Morgan:
Currently, we publish nugets on every commit to release branch,
but gitversion only increments the version if a tag is made
if a tag isn't made, then only the first run of the CI will publish a nuget, otherwise each commit will try and publish a nuget with a semver that already exists.

---

This isn't an ideal situation, since right now, there's no nice way for me to trigger nugets without also triggering a connectors release.
This will cause major issues for managing the nugets on this repo (IFC importer)
We'll need to come up with a better worfklow,
but in the meantime, this PR will remove the unintended behaviour from the current system.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -71,8 +71,9 @@ jobs:
         with:
           name: output-*
 
-  build-linux:
+  deploy-nugets:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


adamhathcock (approved):
Approved minimal change

PR #649, opened 2025-03-05
Build subset when tagged

adamhathcock:
does math between two tags and use affected when patch

found nuget changes too:
image


.config/dotnet-tools.json
```diff
@@ -9,10 +9,10 @@
       ],
       "rollForward": false
     },
-    "gitversion.tool": {
```
oguzhankoral, line 12: dont we need `gitversion` anymore?

Build/Versions.cs
```diff
@@ -0,0 +1,77 @@
...
+      return s_currentFileVersion;
+    }
+    var currentVersion = await ComputeVersion();
+    s_currentFileVersion = currentVersion.WithoutPrereleaseOrMetadata() + ".0";
+    return s_currentFileVersion;
+  }
+
+  public static async Task<string> GetPreviousTag(string currentTag)
+  {
+    var (lastTag, _) = await ReadAsync("git", $"describe --abbrev=0 --tags {currentTag}^");
```
oguzhankoral, line 54: does this consider the versioned tags? as use case scenario, if someone push a tag for any reason to anchor a commit, we could get the wrong state by any chance?

oguzhankoral (changes requested):

Build/Versions.cs
adamhathcock: a tag is a tag....we parse it and it fails the build

.config/dotnet-tools.json
adamhathcock: Nope, we don't use it at all.  At best it was a convenance to get the current tag but that's little of what we need.

PR #661, opened 2025-03-10
test: add tests for receive operation

adamhathcock:


Sdk/Speckle.Connectors.Common/Operations/ReceiveOperation.cs
```diff
@@ -50,15 +49,12 @@ CancellationToken cancellationToken
       .ConfigureAwait(false);
 
     cancellationToken.ThrowIfCancellationRequested();
-    await apiClient.Version.Received(
-      new(version.id, receiveInfo.ProjectId, receiveInfo.SourceApplication),
-      cancellationToken
-    );
+    await receiveVersionRetriever.VersionReceived(account, version, receiveInfo, cancellationToken);
```
JR-Morgan, line 52: Having yet another layer of abstraction around client functions is not ideal....
Is it possible to to moq a `client` or the `VersionResource`...

There's enough functions that require the `Client` that I think its worth us ensuring we create a nice pattern around unit testing it. Especially seeing as we already interface out the `IClientFactory`

JR-Morgan (changes requested):

Sdk/Speckle.Connectors.Common/Operations/ReceiveOperation.cs
adamhathcock: actually this is still going to be necessary because the client object gives a version resource which can't be mocked

adamhathcock:
Includes https://github.com/specklesystems/speckle-sharp-connectors/pull/717 now

JR-Morgan (approved):
Given a quick smoke test in Rhino8

PR #663, opened 2025-03-11
Remove stackalloc usage as it overflows the stack with large meshes

adamhathcock:
Remove stackalloc usage as it overflows the stack with large meshes


didimitrie (approved):
tested locally, can't reproduce my prev fail

JR-Morgan (approved):

oguzhankoral (approved):

PR #669, opened 2025-03-12
fix(Civil3D): Update CorridorDisplayValueExtractor.cs to fix PropertySetDefinitionName error

jhdempsey86:
## Description

For unknown reasons, sometimes trying to get the PropertySetDefinitionName throws an exception. Updated to catch this, ignore that property set, and move to the next one. It is never the "Corridor Identity" property set that has this issue, so there should be no missing data.

## User Value

Stops the conversion from failing due to an erroneous property set whose data is not stored wit the corridor

## Changes:

Caught error when trying to access PropertySetDefinitionName property and continues to next property set.

## To-do before merge:

None

## Screenshots:

image

## Checklist:

- [ X ] My commits are related to the pull request and do not amend unrelated code or documentation.


clairekuang:
Hey Jon, looks fine to me. Just make sure you update the formatting issue so your pr passes our ci!

clairekuang (changes requested):
formatting on 141

clairekuang (approved):

PR #671, opened 2025-03-14
(feat/fix) Navisworks: Add selection progress and try/finally for exceptions

adamhathcock:
image

Added progress for Navisworks selection.
Use the finally block on Navis sending to properly cancel when exceptions happen


jsdbroughton (approved):
Yup

oguzhankoral (approved):
Tested and looks good

PR #672, opened 2025-03-14
Caches visibility for the lifetime of the send

adamhathcock:
requires https://github.com/specklesystems/speckle-sharp-connectors/pull/671

Gets a big speed improvement with a selection that's roughly the same:
image

image


jsdbroughton (approved):
In time if this is a pattern ( i think there are other instances post conversion like it ) perhaps the DI enthusiast would make a CacheMakignServiceFactory

but good stuff for now

PR #599, opened 2025-02-18
fix(autocad/Rhino): add support for receiving any convertible geometry in dataobject display values

clairekuang:
## Description
Some dataobjects (eg Revit areas) have displayvalues that are non-point/polyline/mesh objects.
We should add support to receive any convertible geometry inside the displayvalue of objects.

Also fixes a bug that wasn't allowing us to recieve certain rhino polycurves in autocad. now receives them as a group when there are multiple segments

## User Value

Receive some objects in Rhino and Autocad that previously couldnt be received.

## Changes:

- autocad converter
- rhinoconverter

## Screenshots:

Before:
RHINO: 
- revit areas: https://latest.speckle.systems/projects/2295cb26a0/models/8463e46246@00351004c0
{60AF8954-F2CD-4FCD-8FE7-5CC1D943C205}

AUTOCAD:
- revit areas: https://latest.speckle.systems/projects/2295cb26a0/models/8463e46246@00351004c0
{74817FCD-AB5E-40A8-B8ED-FC7DDA60BE49}

- rhino polycurves in autocad (missing):
{95BEFFC5-F591-4CB7-B344-533D3436372B}

After:
RHINO:
- revit areas: https://latest.speckle.systems/projects/2295cb26a0/models/8463e46246@00351004c0
{14C68C15-23CB-4065-B519-30753788694A}

AUTOCAD:
- revit areas: https://latest.speckle.systems/projects/2295cb26a0/models/8463e46246@00351004c0
{C0B846F6-5533-47FF-A817-3054E0A0EC7F}
- rhino polycurves in autocad:
{F4DB55D6-1812-4657-9F5B-21EA85FB141C}

## Validation of changes:

- tested receiving revit objects model, which have direct shapes with additional geometries (not point, polyline, or mesh) inside display value: https://latest.speckle.systems/projects/2295cb26a0/models/33bb362eab
autocad:
{2B633E51-3D0C-46D4-9F78-0A1ACB6443BB}
rhino:
{FA84B3BF-F4A6-47CE-BCC7-94EA574F7835}

## Checklist:

- [ ] My commits are related to the pull request and do not amend unrelated code or documentation.
- [ ] I have added appropriate tests.
- [ ] I have updated or added relevant documentation.

PR #676, opened 2025-03-14
Fix(autocad): Remove circular dependency

oguzhankoral:
it is introduced here https://github.com/specklesystems/speckle-sharp-connectors/pull/599 but i would expect this should have been a problem on receive testing which claire have many received screenshots in the PR. Maybe circular dependency introduced after the screenshots 🤷‍♂️

image

Bilal's shed
image


clairekuang:
Adding my pr into this one, since this will break the polycurve conversion logic.

tl;dr:  speckle polycurves need to be handled in 1 of 2 ways. 1 way (raw converter) has the exact same return type as the top level converter. Fixing this by copypasta the raw converter logic in the top converter and deleting the raw converter

clairekuang:
after my merge, checked receive in autocad: 
{BB188CFF-780E-4748-98E1-F248797FA0ED}

clairekuang (approved):
checked receiving in autocad and civil, looks like everything is working fine now

clairekuang:
@oguzhankoral was this meant to be hotfixed to main?

oguzhankoral:
> @oguzhankoral was this meant to be hotfixed to main?

yes... maybe we should try to cut release from dev, and lets see how individual connector releases ci will work or not

PR #680, opened 2025-03-14
Feat(accounts): Add remove account to binding

oguzhankoral:
It is a first pass. And tested via console, it does what it supposed to do.

### DIFF
--- a/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Bindings/BasicConnectorBinding.cs
+++ b/Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Bindings/BasicConnectorBinding.cs
@@ -67,6 +67,8 @@ public class BasicConnectorBinding : IBasicConnectorBinding
 
   public void RemoveModel(ModelCard model) => _store.RemoveModel(model);
 
+  public void RemoveModels(List<ModelCard> models) => _store.RemoveModels(models);
+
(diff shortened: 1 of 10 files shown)


Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Bindings/BasicConnectorBinding.cs
```diff
@@ -67,6 +67,8 @@ ITopLevelExceptionHandler topLevelExceptionHandler
 
   public void RemoveModel(ModelCard model) => _store.RemoveModel(model);
 
+  public void RemoveModels(List<ModelCard> models) => _store.RemoveModels(models);
```
didimitrie, line 70: See corollary comment in FE PR: https://github.com/specklesystems/speckle-server/pull/4203#discussion_r1997193697

tl;dr: do we need this vs. removing models in a for loop?

didimitrie (commented):
Is this about removing models in bulk or around removing accounts, or both?

Connectors/ArcGIS/Speckle.Connectors.ArcGIS3/Bindings/BasicConnectorBinding.cs
didimitrie: got my reply in the other pr

didimitrie (approved):

PR #681, opened 2025-03-17
feat (autocad): region and hatches conversions

KatKatKateryna:
## Description

Send Regions and Hatches from AutoCAD.
Add support for receiving regions natively, or being displayValue of a DataObject.
Add ICurveToHostRaw converter to easier iterate through different types of curves.

Example of sent and received back Regions (top row) and Hatches (bottom row): ToSpeckle (composite Hatch lost), ToHost (Hatch pattern lost)  https://app.speckle.systems/projects/16ce7b208c/models/3d914d747f@6b60b13b41
image

## Checklist:


Converters/Autocad/Speckle.Converters.AutocadShared/ToHost/Raw/RegionHatchToHostRawConverter.cs
```diff
@@ -0,0 +1,153 @@
...
+  )
+  {
+    _curveConverter = curveConverter;
+    _settingsStore = settingsStore;
+  }
+
+  public ADB.Hatch Convert(SOG.Region target)
+  {
+    // Get the current document and database
+    Document acDoc = _settingsStore.Current.Document;
```
clairekuang, line 27: Let's try to keep transactions consistent: refer to https://github.com/specklesystems/speckle-sharp-connectors/blob/dev/Converters/Autocad/Speckle.Converters.AutocadShared/ToHost/Geometry/MeshToHostConverter.cs for an example of using the top level transaction instead of creating a new one.

Also maintain naming conventions, because the typical autocad docs examples don't follow our variables. Looks like this code is a bit generated, check this block for an example of opening the block table record for write directly: 
```// append mesh to blocktable record - necessary before adding vertices and faces
    var btr = (ADB.BlockTableRecord)
      tr.GetObject(_settingsStore.Current.Document.Database.CurrentSpaceId, ADB.OpenMode.ForWrite);
    btr.AppendEntity(mesh);
    tr.AddNewlyCreatedDBObject(mesh, true);
```
can replace your lines 27-48 + the intialize hatch method

Converters/Autocad/Speckle.Converters.AutocadShared/ToHost/Raw/RegionHatchToHostRawConverter.cs
```diff
@@ -0,0 +1,153 @@
...
+    Exception? exception = null;
+    try
+    {
+      ConvertAndAssignHatchLoop(acBlkTblRec, acTrans, acHatch, target.boundary, ADB.HatchLoopTypes.External);
+      foreach (var loop in target.innerLoops)
+      {
+        ConvertAndAssignHatchLoop(acBlkTblRec, acTrans, acHatch, loop, ADB.HatchLoopTypes.Outermost);
+      }
+    }
+    catch (Exception ex) when (!ex.IsFatal())
```
clairekuang, line 61: If you're using the top level transaction, you don't need to commit it within the conversion block.
Not sure why you are catching the exception here, only to rethrow later? If you catch the exception, you should be handling it in some way (returning null hatch eg) instead of just rethrowing it later

Converters/Autocad/Speckle.Converters.AutocadShared/ToHost/Raw/RegionHatchToHostRawConverter.cs
```diff
@@ -0,0 +1,153 @@
...
+      {
+        // for simple polylines ".ToPolylineWithPrecision" distorts the shape, so just applying a list of vertices
+        dbCurve = new ADB.Polyline3d(ADB.Poly3dType.SimplePoly, spline.NurbsData.GetControlPoints(), true);
+      }
+      else
+      {
+        dbCurve = spline.ToPolylineWithPrecision(10, false, false);
+      }
+    }
+    ADB.ObjectIdCollection tempDBObjColl = CreateTempObjectIdCollection(acBlkTblRec, acTrans, dbCurve);
```
clairekuang, line 104: check that you don;t need to manually dispose of created objects here afterwards

Converters/Autocad/Speckle.Converters.AutocadShared/ToHost/Raw/RegionToHostRawConverter.cs
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        return adbRegion;
+      }
+    }
+
+    throw new ConversionException($"Region conversion failed: {target}");
+  }
+
+  private ADB.Curve ValidateCurve(List<(ADB.Entity, Base)> convertedResult)
```
clairekuang, line 80: these two methods can just be one

Converters/Autocad/Speckle.Converters.AutocadShared/ToSpeckle/Geometry/HatchToSpeckleConverter.cs
```diff
@@ -0,0 +1,119 @@
...
+    if (loop.IsPolyline)
+    {
+      // disposable object, wrapping into "using"
+      using (AG.Point3dCollection vertices = new())
+      {
+        // collect vertices and construct a polyline simultaneously, it will be clear what to use after iterating
+        ADB.Polyline polyline = new() { Closed = true };
+
+        int count = 0;
+        foreach (ADB.BulgeVertex bVertex in loop.Polyline)
```
clairekuang, line 90: this bulge vertex may need to be manually disposed as well - double check this

Converters/Autocad/Speckle.Converters.AutocadShared/ToSpeckle/Geometry/RegionToSpeckleConverter.cs
```diff
@@ -28,6 +44,106 @@ public SOG.Mesh Convert(ADB.Region target)
...
+        // convert segments to Speckle Polycurve or Circle
+        var convertedLoop = ConvertSegmentsToICurve(segments);
+        loops.Add(convertedLoop);
+      }
+    }
+
+    return loops;
+  }
+
+  private ICurve ConvertSegmentsToICurve(List<AG.Curve3d> segments)
```
clairekuang, line 110: Should this be a separate raw converter?

clairekuang (changes requested):

Converters/Autocad/Speckle.Converters.AutocadShared/ToSpeckle/Geometry/RegionToSpeckleConverter.cs
KatKatKateryna: I woudn't make it a separate converter, because the logic is very much tied to it being one of the Region loops: it converts always to a closed polyline, with only a limited types of supported segments

Converters/Autocad/Speckle.Converters.AutocadShared/ToSpeckle/Geometry/HatchToSpeckleConverter.cs
KatKatKateryna: BulgeVertex is not IDisposable - all good here

Converters/Autocad/Speckle.Converters.AutocadShared/ToHost/Raw/RegionHatchToHostRawConverter.cs
KatKatKateryna: fixed

Converters/Autocad/Speckle.Converters.AutocadShared/ToHost/Raw/RegionHatchToHostRawConverter.cs
KatKatKateryna: super nice, didn't see that we could use it from settingsStore

clairekuang (approved):

PR #682, opened 2025-03-17
refactor: linked model helper class and refactor

bjoernsteinhagen:
## Description

Refactor how linked models are handled. The major changes focus on processing linked models more efficiently (no unnecessary unpacking and processing of elements) and creating a clearer separation of responsibilities between components (got blurry between `RevitSendBinding` and `RevitRootObjectBuilder`).

## User Value

Would significantly improves performance when sending models with disabled linked model setting, especially for large projects with many linked models.

###  Premature Processing of Linked Models ✅ FIXED
Addressed in `RevitSendBinding.cs`.
- Early check of linked model setting
- See comments for better explanation
- Only collects linked model elements IF setting is enabled

### Mixing of Concerns ✅ FIXED
Getting some clearer separation of responsibilites between the three classes:
- `LinkedModelHandler` - this is a new class. Dedicated to element collection mechanics from linked models and handling different filter types (categories vs. selection) ONLY. Won't respond to UI setting for linked models.
- `RevitSendBinding` - controls flow and handles decision-making based on UI setting.
- `RevitRootObjectBuilder` - processes contexts and delegates to conversions. We have maintained the warning users get if a linked model is present but no setting enabled => this approach (to signal that warnings are needed) relies on empty element lists sent from `RevitSendBinding` which smells and is a bit of an implicit mechanism (but temp. fix and works for now).

### Selection including linked models but setting not enabled:

image

#### Main model + linked models

image

#### Main model only

image

## Checklist:

### DIFF
--- a/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
+++ b/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
@@ -232,6 +232,8 @@ internal sealed class RevitSendBinding : RevitBaseBinding, ISendBinding
           linkedDocumentContexts.Add(new(transform, linkedDoc, linkedElements));
         }
         // ⚠️ when disabled, still adds empty contexts to maintain warning generation in RevitRootObjectBuilder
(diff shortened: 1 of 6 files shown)


oguzhankoral (approved):
Nice! There are still some BUTs, but okay for now.

PR #688, opened 2025-03-19
fix(revit): handle multiple linked model instances with transform hashing

bjoernsteinhagen:
## Description

Add transform-specific hashing to properly handle multiple instances of the same linked model. Uses transform properties to create a unique suffix for the applicationId, preventing cached objects from being incorrectly reused across different transform contexts. This also fixes problems observed when trying to recieve back into Revit.

Fixes:
- [CNX-1385](https://linear.app/speckle/issue/CNX-1385/linked-models-instancing)
- [CNX-1434](https://linear.app/speckle/issue/CNX-1434/loading-linked-models-back-in-revit).

## User Value

Lots.

## Changes:

- Added a `GetTransformHash` method to create unique hashes for transforms (based on `IdGenerator.cs` from `Base`)
- Modified caching logic in `RevitRootObjectBuilder` to consider element transformations
- Added unique transform-specific suffixes to `applicationId` for elements with transforms
- If `revitElement` doesn't have a transform, then its `applicationId` just stays as is

## Screenshots:

You can see the Speckle model here: [cnx-1385](https://latest.speckle.systems/projects/99d30f721c/models/314fb22e94)

### Same wall instances suffixed with transform hash

image

image

### Main model elements remain unchanged

image

## Validation of changes:
- We send a model with **mutiple instances of the same linked model**
- A wall from the main model is shown with a normal looking `applicationId`
- A wall from an instance of a linked model has an `applicationId` with `_t` suffix and the `transformHash`
- This model can then be received back in Revit with no issues, fixing [CNX-1434](https://linear.app/speckle/issue/CNX-1434/loading-linked-models-back-in-revit).

https://github.com/user-attachments/assets/b976b99b-ac6a-4b2c-b2f2-8fd3e5c67680

## Checklist:

### DIFF
--- a/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/RevitRootObjectBuilder.cs
+++ b/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/RevitRootObjectBuilder.cs
@@ -12,6 +12,7 @@ using Speckle.Converters.Common;
 using Speckle.Converters.RevitShared.Helpers;
 using Speckle.Converters.RevitShared.Settings;
 using Speckle.Sdk;
+using Speckle.Sdk.Common;
 using Speckle.Sdk.Models;
 using Speckle.Sdk.Models.Collections;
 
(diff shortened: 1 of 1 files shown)


oguzhankoral (commented):
I would test it with cache invalidation -means that after first send, move one of the linked models and send again. check moved objects whether in correct locations or not

oguzhankoral (approved):

PR #690, opened 2025-03-20
feat(grasshopper): add property receive nodes

clairekuang:
## Description

Adds two components:
- Property Group Path Selector
- Filter Object Properties by Path

This includes 2 new goos/params:
- PropertyGoo
- PropertyGroupGoo

## Screenshots:

{E11B399C-603E-4278-9724-5BD7F77E5EC7}


clairekuang:
Ok, made an update to use hashsets. This is the current output for multiple input objects:
{7FC62058-D5D3-4158-8361-296622B34A3E}

PR #696, opened 2025-03-24
feat (revit): receive Region as native FilledRegion

KatKatKateryna:
## Description
Needs opinionated decision! Detailed investigation here: https://linear.app/speckle/issue/CNX-1363/add-fallback-receive-converters-for-all-remaining-c-connectors-revit
This PR adds RegionToHost converter for Revit. The main challenge is that native Revit class for CAD Regions and Hatches is FilledRegion (this is what is being created for native CAD imports). But this class is not what our BaseToHost converter expects (it expects GeometryObject), therefore there are some changes to accept just a string as a conversion result (element unique ID). Also, it can only be drawn in a suitable 2d View and will fail in a 3d View, this is why if ONLY Regions are being converted while in 2d View, the Model Highlight feature will not detect any objects (because it defaults to a 3d View). 

Proposed solution follows the way native CAD import works: the converter type depends on the Active View. If Active View is 2d, then Filled Region will be created (but untraceable in our report). If ActiveView is 3d, then the converter will default to Mesh converter and be traceable in the Report.

Commits to test with: https://app.speckle.systems/projects/16ce7b208c/models/e9f8edeb13@2637957c82 , https://latest.speckle.systems/projects/2295cb26a0/models/874439f06c@a5d9165893

Example of a native FilledRegion receive, with the limitations that user should make sure things are visible in the current view: 
image (1)

Example side-by-side receiving data in 3d view, then switching to 2d, slightly moving and receiving again:
image

## Changes:
The only seemingly unrelated change to RegionToHost converter, is removal of empty list check in RevitRootToHost Converter. This check is redundant (called directly after BaseToHost, which is already doing the same check). Also, it prevented us from making an exception and returning empty list in case of Regions. 

## Checklist:


Converters/Revit/Speckle.Converters.RevitShared/RevitRootToHostConverter.cs
```diff
@@ -28,11 +27,6 @@ public object Convert(Base target)
   {
     List<GeometryObject> geometryObjects = _baseToGeometryConverter.Convert(target);
 
-    if (geometryObjects.Count == 0)
```
KatKatKateryna, line 31: removed redundant check (already done in BaseToHost converter)

Converters/Revit/Speckle.Converters.RevitShared/RevitRootToHostConverter.cs
```diff
@@ -13,24 +13,69 @@ public record DirectShapeDefinitionWrapper(string DefinitionId, List<GeometryObj
...
   {
-    List<GeometryObject> geometryObjects = _baseToGeometryConverter.Convert(target);
+    List<GeometryObject> geometryObjects = new();
+    List<string> regionIds = new();
 
-    if (geometryObjects.Count == 0)
+    switch (target)
     {
-      throw new ConversionException($"No supported conversion for {target.speckle_type} found.");
+      case SOG.Region region:
```
clairekuang, line 37: The root to host converter is not the right place to place this logic. Just like there's a base to geometry converter, you should instead add a base to annotation converter that has the region converter.
 You may consider chekcing for active view, to determine whether to use the annotation converter vs geometry converter

clairekuang (approved):
The changes I suggested aren't essential, but do make the code cleaner by separating responsibility for view management and the host object builder. Up to you if you want to make the changes.

Converters/Revit/Speckle.Converters.RevitShared/RevitRootToHostConverter.cs
```diff
@@ -14,19 +14,31 @@ public class RevitRootToHostConverter : IRootToHostConverter
...
   {
+    _planViewToGeometryConverter = planViewToGeometryConverter;
     _baseToGeometryConverter = baseToGeometryConverter;
     _converterSettings = converterSettings;
   }
 
   public object Convert(Base target)
   {
-    List<GeometryObject> geometryObjects = _baseToGeometryConverter.Convert(target);
+    // If ActiveView is a vertical 2d view, use PlanView converter (will ignore DirectShapes)
```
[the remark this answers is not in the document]
KatKatKateryna: Adjusted the HostObjectBuilder by injecting ViewManager, but it will not be accessible in Converters project

KatKatKateryna:
I noticed an interesting side effect: when receiving in 2d View, any model card (added before or after) will jump to this 2d view on HighlightModel or Object

Connectors/Revit/Speckle.Connectors.RevitShared/HostApp/RevitViewManager.cs
```diff
@@ -0,0 +1,48 @@
...
+      case ViewType.Detail:
+        return IsHorizontalView(activeView);
+
+      default:
+        return false;
+    }
+  }
+
+  public bool Is3dView(View activeView)
+  {
```
clairekuang, line 32: this can be simplified to:
```
public bool Is3dView(View activeView) => activeView.ViewType == ViewType.ThreeD;
```

Connectors/Revit/Speckle.Connectors.RevitShared/HostApp/RevitViewManager.cs
```diff
@@ -0,0 +1,48 @@
...
+    if (activeView.ViewType == ViewType.ThreeD)
+    {
+      return true;
+    }
+    return false;
+  }
+
+  private bool IsHorizontalView(View activeView)
```
clairekuang, line 40: similarly:
```
bool IsHorizontalView(View activeView) => Math.Abs(activeView.ViewDirection.Z - 1)< 0.00001 ;

clairekuang (approved):

PR #712, opened 2025-03-25
fix(ifc): Fixed regression with IFC Site geometry not being converted

JR-Morgan:
See ticket


Importers/Ifc/Speckle.Importers.Ifc/Converters/SpatialStructureConverter.cs
```diff
@@ -1,31 +1,52 @@
...
+    // 1. Spatial elements can can have direct geometry (mostly only common with IFC Site)
+    // 2. Keeps property access simpler
     return new Collection
     {
-      name = node.Name ?? node.Guid,
-      applicationId = node.Guid,
-      elements = childrenConverter.ConvertChildren(model, node),
+      name = node.Name ?? node.LongName ?? node.Guid,
+      elements = allChildren,
+      ["expressID"] = node.Id,
```
didimitrie, line 28: would even remove this to prevent two objects having the same express id, but not a blocker

didimitrie (approved):

PR #718, opened 2025-03-27
Reworded error message for DataObjects that have no displayValue

JR-Morgan:
re-name the exception message from "Failed to convert object"
to "Object did not convert to any native geometry"

I'm seeing this error a lot when receiving DataObjects with no displayValues (from IFC)

Its still not really a failure, but at least the exception message is more descriptive


Connectors/Rhino/Speckle.Connectors.RhinoShared/Operations/Receive/RhinoHostObjectBuilder.cs
```diff
@@ -187,7 +187,7 @@ CancellationToken cancellationToken
             if (conversionIds.Count == 0)
             {
               // TODO: add this condition to report object - same as in autocad
-              throw new SpeckleException($"Failed to convert object.");
```
oguzhankoral, line 190: did you check messaging of other connectors? curious why you only interested with rhino for ifc

JR-Morgan: I didn't rigorously test this in all connectors, but only rhino appears to have such a check.

oguzhankoral (approved):

PR #726, opened 2025-03-30
feat(rhino): receives properties as user strings

clairekuang:
## Description

Enables receiving properties as user strings, where all nested property dicts will be flattened to a `key.key.key` etc format.
All values are converted to strings, null values will be an empty string

## User Value

enables coordination workflows using properites

## Changes:

- rhino connector

## Screenshots:
{C7937871-A328-4E93-86B7-A1967CEE8756}

## Validation of changes:
 tested receiving revit objects

## Checklist:


didimitrie (approved):

oguzhankoral (approved):
Tested and havent feel performance slow down, looks good!

PR #727, opened 2025-04-01
fix(revit): Enable view-based filtering for multiple linked model instances

bjoernsteinhagen:
## Description

This PR fixes an issue with the view-based filter in Revit where multiple instances of the same linked model were not being handled correctly. Previously, only the visibility of the first linked model instance was considered when determining which elements to include. This meant that if a user had multiple instances of the same linked model with different visibility settings in a view, the system would incorrectly use only the visibility of the first instance for all instances.

The fix leverages our existing transform hash mechanism to identify and match specific linked model instances, ensuring that each instance's visibility in the current view is correctly evaluated.

Fixes [CNX-1490](https://linear.app/speckle/issue/CNX-1490/section-boxes-and-linked-models)

## User Value

Users can now accurately send elements from specific instances of linked models based on their visibility in the current view, providing more precise control over which elements are included in a Speckle send.

## Changes:

- Changes to `LinkedModelHandler.FindLinkInstanceForDocument` to identify specific linked model instances based on transform hash
- Updated `GetLinkedModelElements` to pass transform information for proper instance matching

## Screenshots:

Screenshot 2025-04-01 094231

image

## Validation of changes:

- Sending linked models with multiple instances. Testing if the instances visibility on send are independant of the first instance.

## Checklist:

### DIFF
--- a/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
+++ b/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
@@ -233,7 +233,7 @@ internal sealed class RevitSendBinding : RevitBaseBinding, ISendBinding
         if (includeLinkedModels)
         {
           // handler is only responsible for element collection mechanics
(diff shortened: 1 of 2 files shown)


oguzhankoral (approved):
Looks legit!

PR #740, opened 2025-04-08
fix(civil3d): Fix further alignment spiral/arc/line bug

jhdempsey86:
## Description

Following on from https://github.com/specklesystems/speckle-sharp-connectors/pull/692/files, we found another test model where getting the MinimumTransitionLength of a spiral threw the same InvalidOperation error. 

Checking the documentation, nearly all properties on the AlignmentSubEntityLine, AlignmentSubEntityArc, and AlignmentSubEntitySpiral objects can throw an InvalidOpearation exception when you try to access them.

Therefore, to stop Speckle failing to convert an object due to any of these throwing an error, getting the value now goes through a TryGetValue method that will return null if accessing the property throws an InvalidOPerationException.

## User Value

Stops Speckle failing to convert the alignment if any of the properties throw an exception. It means more successfull uploads!

## Changes:

Below are the changes to the AlignmentSubentitySpiralToSpeckleRawConverter class
Similar changes have been made to the AlignmentSubentityLineToSpeckleRawConverter and AlignmentSubentityArcToSpeckleRawConverter classes

image

## To-do before merge:

At the moment, I have duplicated the TryGetValue method in each of the three classes because i was not sure where it might be best to put this function to make it available to all 3 classes that are using it.
Please feel free to move the location of this method to somewhere more suitable to avoid duplication

## Screenshots:

Before:
image

After:
image

## Validation of changes:

Test model which throws the error now successfully uploads all the alignments within it

## Checklist:

- [X] My commits are related to the pull request and do not amend unrelated code or documentation.
- [X] I have added appropriate tests.
- [X] I have updated or added relevant documentation.


clairekuang:
@jhdempsey86 happy to approve this after you fix the CI build format issue, but I will probably move this logic out to its own propertyHandler class, since these situations happen in various places in the converter. This way we can avoid copypasted methods

clairekuang (changes requested):
just a formatting issue in the catch block. I recommend installing csharpier and changing your settings to autoformat on save, to make it easier for future dev

clairekuang (approved):
Pushed formatting fixes to your branch. Merging!

PR #743, opened 2025-04-08
refactor(ci): Update workflow to use new consolidated deployment workflow

JR-Morgan:
see https://linear.app/speckle/issue/CXPLA-177/clean-up-connector-installer-ci

Before this PR can be merged, several other PRs need to be merged at the same time. Let @JR-Morgan orchestrate merging.


.github/workflows/release.yml
```diff
@@ -35,30 +37,38 @@ jobs:
...
-          inputs: '{ "run_id": "${{ github.run_id }}", "version": "${{ needs.build-windows.outputs.version }}", "public_release": ${{ env.IS_PUBLIC_RELEASE }}, "store_artifacts": ${{ env.IS_TEST_INSTALLER }} }'
-          ref: main
+          inputs: '{
+            "run_id": "${{ github.run_id }}",
+            "semver": "${{ needs.build-windows.outputs.semver }}",
+            "file_version": "${{ needs.build-windows.outputs.file_version }}",
+            "repo": "${{ github.repository }}",
+            "is_public_release": ${{ env.IS_PUBLIC_RELEASE }}
+            }'
+          ref: jrm/experiment-similar-ci
```
adamhathcock, line 71: remove/change?

JR-Morgan: Yep I'll change them all to `main`

.github/workflows/release.yml
```diff
@@ -3,15 +3,17 @@ name: .NET Build and Publish
 on:
   push:
     branches: ["main", "installer-test/**"]
-    tags: ["v3.*"] # Manual delivery on every 3.x tag
+    tags: ["v3.*.*"] # Manual delivery on every 3.x tag
```
oguzhankoral, line 6: hmm, what is changing here as behavior?

JR-Morgan: Some repos were using `v3.*`, others were using `v3.*.*`.

I wanted to align everything, so I picked the one that is least likely to lead to an invalid semver

JR-Morgan: before. a tag of `v3.1` would match, and trigger..
now you need to do atleast two `.` e.g. `v3.1.0` or `v3.1.0-beta` etc...

oguzhankoral: oki, make sense

PR #744, opened 2025-04-08
feat(Autocad 2026) Add Autocad and Civil3d 2026

adamhathcock:


oguzhankoral (approved):
Tested both autocad and civil for send and receive. all looks good

PR #763, opened 2025-04-18
Fix (revit): disable regions in revit

oguzhankoral:
Now we can receive sketchup commits, and regions as fallback mesh

image


bimgeek (approved):
Tested together with Ogu. Seems to resolve the previous issue of Sketchup not being received in Revit. It's also consistent with how we receive other elements.

PR #770, opened 2025-04-21
fix: allows for stacked walls to be sent

didimitrie:
Stacked walls now send. See various attempts here: https://app.speckle.systems/projects/0faf9cbcca/models/9e5bef2441/versions

Tested in revit 2022 and 2023.


oguzhankoral (approved):
Tested too

PR #796, opened 2025-04-30
ci(grasshopper): Grasshopper Installer 

JR-Morgan:
Adds necessary changes for Grasshopper Deployment.

### Fix 1: Resource Files
Previously, the namespace of the built resource file was being pulled from the Grasshopper8 project, and thus was causing exceptions for the Grasshopper7 connector failing to resolve the resources.

To fix this, All projects importing the Grasshoppper shared project, need to set their root namespace to align with the shared project.

### Fix 2: DLL conflicts between Grasshopper and Rhino connectors:
Previously, the Grasshopper and Rhino projects could be built indipendently, but because of differences in their dependencies, they would be built with different versions of dependencies (`System.Memory`, `System.Numerics.Vectors`, `System.Runtime.CompilerServices.Unsafe` and `System.Buffers`)

Rhino documents now reference the Grasshopper projects. This is the same as in v2, and is needed to ensure we build 1 project with no dependency conflicts. At the end of the day, all the files get copied to the same folder. This was blocking side by side Grasshopper and Rhino connectors working.

Because of this, we don't need to add the grasshopper projects to the build consts. We grab the files from the Rhino build, and dotnet affected sees the project reference and will beave correctly.

image

### Cleanup: 
There was some crap in the csproj for Grasshopper projects, did some light cleanup


adamhathcock (approved):
Looks good but there's something to revert!

Connectors/Rhino/Speckle.Connectors.Rhino7/Speckle.Connectors.Rhino7.csproj
```diff
@@ -18,8 +18,7 @@
   </ItemGroup>
  
   <ItemGroup>
-    <ProjectReference Include="..\..\..\Converters\Rhino\Speckle.Converters.Rhino7\Speckle.Converters.Rhino7.csproj" />
-    <ProjectReference Include="..\..\..\Sdk\Speckle.Connectors.Common\Speckle.Connectors.Common.csproj" />
+    <ProjectReference Include="..\..\..\Connectors\Rhino\Speckle.Connectors.Grasshopper7\Speckle.Connectors.Grasshopper7.csproj" />
```
adamhathcock, line 21: wait, so Rhino without Grasshopper isn't a thing?

JR-Morgan: Correct, this is the way we did it in v2. While Rhino doesn't rely on the grasshopper code. we want to make sure it is picking dlls that don't conflict in anyway.

Seeing as we'll always be shipping rhino with grasshopper, this was an easy option. But if you can think of alternatives, happy to consider.

oguzhankoral (approved):

PR #798, opened 2025-04-30
Generate HostApp specific sln and slnx 

adamhathcock:
Add SLNX files and generate host app specifc ones.  README updated

The new SLNX files for each connector similar to SLNF had in v2
Unlike the SLNF files, we don't need to manually maintain them, they are source generated from the main SLN/SLNX
Really important that people don't try and modify them, since they're source generated.
SLNX will eventually replace SLN standard, but it's not fully adopted yet. to open them you need Rider 2024.3 , or Visual Studio 17.13

https://linear.app/speckle/issue/CNX-1182/redo-solutions-to-be-allsln-with-filters-for-things


JR-Morgan (approved):
Can't wait for SLN to be replaced by SLNX.

PR #801, opened 2025-05-01
skip converting breps/mehses with skew transform

kekesidavid:
temporary fix for:
https://linear.app/speckle/issue/CNX-1671/rhino2revit-block-transformations-are-broken-again

the fix will throw:
throw new SpeckleException("Transform was not conformal. Skew, non uniform scale is currently not supported.");
if we find a transform with skew

normalization of basis vectors is removed from the transform
applying the transforms is reversed

we will skip the highlighted Breps/Meshes for now:
image

### DIFF
--- a/Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/LocalToGlobalToDirectShapeConverter.cs
+++ b/Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/LocalToGlobalToDirectShapeConverter.cs
@@ -1,4 +1,4 @@
-﻿using Speckle.Converters.Common;
+using Speckle.Converters.Common;
 using Speckle.Converters.Common.Objects;
 using Speckle.Converters.RevitShared.Settings;
 using Speckle.DoubleNumerics;
@@ -75,7 +75,7 @@ public class LocalToGlobalToDirectShapeConverter
     // existence of units is must, to be able to scale the transform correctly
     if (target.atomicObject["units"] is string units)
     {
-      foreach (Matrix4x4 matrix in target.matrix)
+      foreach (Matrix4x4 matrix in target.matrix.Reverse())
       {
         DB.Transform revitTransform = _transformConverter.Convert((matrix, units));
         combinedTransform = combinedTransform.Multiply(revitTransform);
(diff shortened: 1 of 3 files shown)


bimgeek (approved):
LGTM

Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/LocalToGlobalToDirectShapeConverter.cs
```diff
@@ -75,7 +75,7 @@ public DB.DirectShape Convert((Base atomicObject, IReadOnlyCollection<Matrix4x4>
     // existence of units is must, to be able to scale the transform correctly
     if (target.atomicObject["units"] is string units)
     {
-      foreach (Matrix4x4 matrix in target.matrix)
+      foreach (Matrix4x4 matrix in target.matrix.Reverse())
```
oguzhankoral, line 78: Did we test any side effect on real models instead sample object model?

kekesidavid: i only tested with this model: https://latest.speckle.systems/projects/58241c04d1/models/73412a6ecf

oguzhankoral: ok this is not enough, we should test it with real models i.e. revit sample school, some complex rhino files etc

kekesidavid: i agree, but ask Bilal what models he used for testing

oguzhankoral: i didn't get it, PR author should also be testing by default, the code pieces you changed can have many random side effects on different scenarios

bjoernsteinhagen:
> LGTM

Skipping objects that don't convert properly :confused: Are we comfortable doing this because these objects occur so rarely?

Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/LocalToGlobalToDirectShapeConverter.cs
bjoernsteinhagen: I'll pick up testing 👍🏼

oguzhankoral (approved):

PR #806, opened 2025-05-02
fix(revit): publish id instead of name for type id

bjoernsteinhagen:
## Description

Related to [CNX-1125](https://linear.app/speckle/issue/CNX-1125/publish-type-id-instead-of-name).

## User Value

Requested [here](https://speckle.community/t/finding-type-ids-for-revit-elements/16117/3). 

## Changes:

- Explicit check for `Type ID` parameter and subsequent setting of `id`

### Before

image

### After

Commit after changes - [model](https://app.speckle.systems/projects/03074a2834/models/1f93f2e3b6).

image

## Validation of changes:

See screenshots.

## Checklist:

### DIFF
--- a/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/Properties/ParameterExtractor.cs
+++ b/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/Properties/ParameterExtractor.cs
@@ -133,7 +133,10 @@ public class ParameterExtractor
           continue;
         }
 
-        var value = GetValue(parameter);
+        // "SYMBOL_ID_PARAM" is internal name for "Type ID". localization impacts definition names, internal more reliable
+        // probably a waste to reuse HandleDefinition method here? simple approach would be to change method signature for GetValue to also take
+        // internalDefinitionName as argument. we're discarding most returns here anyway
+        object? value = internalDefinitionName == "SYMBOL_ID_PARAM" ? parameter.AsElementId() : GetValue(parameter);
 
         var isNullOrEmpty = value == null || (value is string s && string.IsNullOrEmpty(s));
 
@@ -218,15 +221,6 @@ public class ParameterExtractor
         if (elId == DB.ElementId.InvalidElementId)
         {
           return null;
-        }
-
-        // "SYMBOL_ID_PARAM" is internal name for "Type ID". localization impacts definition names, internal more reliable
-        // probably a waste to reuse HandleDefinition method here? simple approach would be to change method signature for GetValue to also take
-        // internalDefinitionName as argument. we're discarding most returns here anyway
-        var (internalDefinitionName, _, _, _) = _parameterDefinitionHandler.HandleDefinition(parameter);
-        if (internalDefinitionName == "SYMBOL_ID_PARAM")
-        {
-          return elId.ToString();
         }
 
         if (_elementNameCache.TryGetValue(elId, out object? value))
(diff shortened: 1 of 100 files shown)


Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/Properties/ParameterExtractor.cs
```diff
@@ -218,6 +219,13 @@ ILogger<ParameterExtractor> logger
           return null;
         }
 
+        // NOTE: first check if this is a "Type ID" parameter, if so return the actual ID
+        // explicit (aka hacky) check here related to [CNX-1125](https://linear.app/speckle/issue/CNX-1125/publish-type-id-instead-of-name)
+        if (parameter.Definition.Name.Equals("Type ID", StringComparison.OrdinalIgnoreCase))
```
clairekuang, line 224: confirming that localization doesn't impact definition names? is internal name more reliable?

bjoernsteinhagen: ooohpf. good catch! updated in `fix: internal definition name` to use internal name and tested with other language:

image

Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/Properties/ParameterExtractor.cs
```diff
@@ -218,6 +220,15 @@ ILogger<ParameterExtractor> logger
           return null;
         }
 
+        // "SYMBOL_ID_PARAM" is internal name for "Type ID". localization impacts definition names, internal more reliable
+        // probably a waste to reuse HandleDefinition method here? simple approach would be to change method signature for GetValue to also take
+        // internalDefinitionName as argument. we're discarding most returns here anyway
+        var (internalDefinitionName, _, _, _) = _parameterDefinitionHandler.HandleDefinition(parameter);
```
bjoernsteinhagen, line 226: @clairekuang using `HandleDefinition` makes sense to me, but it has already been called in `ParseParameterSet` (which then calls this `GetValue` method).

I'm inclined to change the method signature for `GetValue` to include the `internalDefinitionName` as an arg, so we don't need to recall it here?

clairekuang (approved):

Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/Properties/ParameterExtractor.cs
```diff
@@ -125,13 +125,18 @@ ILogger<ParameterExtractor> logger
...
+        // "Type ID" (associated with "SYMBOL_ID_PARAM") won't evaluate to true here which is intentional
         if (internalDefinitionName.EndsWith("_ID") || internalDefinitionName.EndsWith("_PARAM_ID"))
         {
           continue;
         }
 
-        var value = GetValue(parameter);
+        // NOTE: excepted behaviour is to use GetValue BUT if we have "SYMBOL_ID_PARAM", we just want id
+        // see above comment and linked linear ticket / issue.
+        object? value =
```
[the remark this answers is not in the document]
bjoernsteinhagen: Good question - I put in a `counter` for the number of times we hit a `SYMBOL_ID_PARAM` per `ParameterSet` and it was never more than 1, so I'm confident it's not changing anything else but that one prop on each `element`.

image

PR #808, opened 2025-05-05
Kateryna/cnx 1537 autocad send text

KatKatKateryna:
## Description

Support for sending text from AutoCAD: DBText (1 line only) and MText (Multiline).
Conversions are separated into top-level and raw, because Text conversions will likely be used as raw converters in more complex objects later: e.g. Dimensions etc.

Limitations:
- MText split into columns will ignore the columns, and keep the "width" of all columns together as just a text maxWidth
- Some Justifications are not accounted for (e.g. BaseAlign, BaseFit etc.) because they are not used in the simple MText objects (possibly when they are a part of Dimension or some other objects that we don't support yet).

Other current limitations in the viewer display:

Horizontal and Vertical Alignment points are not supported (defaults to Top Left)
Height: font in pixels (if Units.None) is interpreted as linear units
if Plane is nullable, the text should follow the camera (like Measurements in the viewer) but for now just displays in XY plane starting in Origin Point
maxWidth: maximum width of the text (in the specified units) should split it into lines (null for "don't split"), but currently doesn't make any difference visually
Also I noticed if the color proxies are on the Collection level, the text doesn't pick it up and displays grey

Sample file and Speckle Model: https://drive.google.com/file/d/1pOeBimUTzW9el2DFGhcImKyHGqkZrP8t/view?usp=drive_link 
https://latest.speckle.systems/projects/2295cb26a0/models/7c9dc1dcd8 
image

## Checklist:


clairekuang (approved):
code-wise looks fine to me

PR #812, opened 2025-05-05
Dim/gh props

didimitrie:
Several fixes & improvements: 
- [fix] property creation node is now item based, simplifying a lot of flows and relying on gh for natural data lacing 
- [fix] property path selector now ignores input structure and will always output flat list of prop keys
- [fix] get object property node now works with multiple props, and respects input tree structure (again, swapped to item based input)
- [feat] improves diffing for output port generation for nicer user experience when swapping props in and out (less need to recreate downstream connections as we're keeping as many existing params alive)

Most changes visible in this screenshot: 

image

Closes [CNX-1735: Getting more than two props throws an error (but works)](https://linear.app/speckle/issue/CNX-1735/getting-more-than-two-props-throws-an-error-but-works)


Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Components/Objects/CreateSpeckleProperties.cs
```diff
@@ -61,46 +54,23 @@ protected override void SolveInstance(IGH_DataAccess da)
...
     {
-      var currentProps = new Dictionary<string, SpecklePropertyGoo>();
-      foreach (var kvp in properties)
+      object? value = null;
+      var success = da.GetData(i, ref value);
+      var actualValue = value?.GetType().GetProperty("Value").GetValue(value); // note: unsure if reflection here hurts our performance
+      if (!success || value == null || actualValue == null)
       {
-        currentProps[kvp.Key] = kvp.Value[i];
+        continue;
```
clairekuang, line 66: We probably want to add a flag before continuing, so the user knows their props are a little borked: 
```AddRuntimeMessage(
            GH_RuntimeMessageLevel.Error,
            $"Parameter {inputParam.NickName} should not contain anything other than strings, doubles, ints, and bools."
          );```

PR #813, opened 2025-05-06
fix(revit): ITransformable objects loading in groups and duplicated

bjoernsteinhagen:
## Description

Fixes [CNX-1707](https://linear.app/speckle/issue/CNX-1707/revit-curves-and-meshes-in-blocks-come-as-duplicated) where block instances with `ITransformable` objects (like polylines, points, etc.) behaving incorrectly when loaded in Revit. Same objects in multiple instances were grouped and duplicated n times (n = number of instances of a block definition).

Problem in how the `TransformTo` method implementation copies the `applicationId` when creating transformed instances, leading to multiple DirectShapes sharing the same definition in the DirectShapeLibrary.

## User Value

Clean receive.

## Changes:

- Appending short GUID to in `RevitHostObjectBuilder`. Makes no sense in the context of `applicationId` but ensures a clean receive.

## To-Do Before Merge
@oguzhankoral discuss this hacking of `applicationId`. It smells!

## Validation of changes:

https://github.com/user-attachments/assets/8a10a883-d79b-488f-836e-4f0b571b620e

- Worried about performance implications. Profiled code before and after changes with no noticeable change
- Logs for successfull and unsuccessful conversions before and after remained the same 

## Checklist:

### DIFF
--- a/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Receive/RevitHostObjectBuilder.cs
+++ b/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Receive/RevitHostObjectBuilder.cs
@@ -188,15 +188,21 @@ public sealed class RevitHostObjectBuilder(
           //TODO TransformTo will be deprecated as it's dangerous and requires ID transposing which is wrong!
           //ID needs to be copied to the new instance
           var id = localToGlobalMap.AtomicObject.id;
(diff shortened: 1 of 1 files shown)


oguzhankoral (approved):
Harmless mutation on receive for `applicationId`. Overall, local to global is a hack in revit and we might need to reconsider it sometime

PR #819, opened 2025-05-09
Kateryna/cnx 1762 texts inside dynamic blocks

KatKatKateryna:
## Checklist:


KatKatKateryna:
@clairekuang comments and references added 🙌
@bimgeek the decision was unfortunately to keep the behavior aligned with what AutoCAD API provides, so this will not be a very nice selection behavior in the Viewer (text will not be selected as a part of BlockInstance). The reason is that creating new BlockDefinitions which don't exist in AutoCAD API, might lead to more side effects that we haven't anticipated, because creating custom definitionIDs might be not as harmless as it seems.

clairekuang (approved):

PR #820, opened 2025-05-09
Feat(gh): workspaces and search wizard

oguzhankoral:
Extracts out search and cascaded selection logic outside of GH component to save the next step which we will have this logic in 4 different components.

### When `userSelectedAccountId` is not set yet (if user havent used dui3 or other connectors yet)
Whenever an account set, it remembers later since we set it
image

### Defaulting to `userSelectedAccountId` when component placed
image

### No workspace yet scenario -> navigates to `/workspaces/actions/create` end point for relevant account server
image

### Defaulting to `activeWorkspace` when component placed for last used account
image

### Disabled projects according to `canLoad` and `canPublish` permissions
image

### Versions are limited according to plan
image

### When URL string provided, buttons disabled and info reflected into buttons
image

### Self hosters scenario that doesn't have workspaces
image

### URL does not have enough permission scenario
image


Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Components/Operations/Wizard/SpeckleOperationWizard.cs
```diff
@@ -0,0 +1,324 @@
...
+    _refreshComponent = refreshComponent;
+    _clientFactory = PriorityLoader.Container.GetRequiredService<IClientFactory>();
+    _accountManager = PriorityLoader.Container.GetRequiredService<AccountManager>();
+    _accountService = PriorityLoader.Container.GetRequiredService<AccountService>();
+
+    var userSelectedAccountId = _accountService.GetUserSelectedAccountId();
+    SelectedAccount =
+      userSelectedAccountId != null
+        ? _accountManager.GetAccount(userSelectedAccountId)
+        : _accountManager.GetDefaultAccount();
```
clairekuang, line 52: Let's try to not use the default here, shouldn't be null at this point but throw exception if it is

Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Components/Operations/Wizard/SpeckleOperationWizard.cs
```diff
@@ -0,0 +1,324 @@
...
+
+    WorkspaceMenuHandler = new WorkspaceMenuHandler(FetchWorkspaces, CreateNewWorkspace);
+    ProjectMenuHandler = new ProjectMenuHandler(FetchProjects); // TODO: Nullability of account need to be handled before
```
clairekuang, line 55: I think nullability has been handled? can remove this comment

clairekuang (approved):
Just for safety, bump sdk version to 3.3.3 in dev and make any needed changes, so we won't have to do it on the gh branch

PR #824, opened 2025-05-12
Kateryna/cnx 1615 hatch loop conversion failed

KatKatKateryna:
## Description
Conversion of some hatches and regions ToSpeckle were failing. Reasons: unsupported types of Hatch Loops and Region Loops. 
image

What has been added for Regions: 
- Handling edge case when the boundary (or inner loop) of the Region is a close Ellipse
- Handling the case when the boundary (or inner loop) of the Region is a NurbCurve3d: added a Raw ToSpeckle Converter. NurbCurve3d is very similar to ADB.Spline, so most of the logic was copy-pasted with adjusting the names of properties and methods.

What has been added for Hatches:
- Handling the case when the boundary (or inner loop) doesn't have a Polyline representation. In this case, we recreate the ADB.Polyline segment by segment, extracting the vertices and bulges from each. In some edge cases, was can also return a full circle or ellipse, or a Spline representation of a NurbCurve (which will be the most exact representation).

image

## Checklist:


Converters/Autocad/Speckle.Converters.AutocadShared/ToSpeckle/Geometry/HatchToSpeckleConverter.cs
```diff
@@ -110,10 +110,145 @@ private ADB.Curve PolylineFromLoop(ADB.HatchLoop loop)
             );
           return new ADB.Circle(centerPt, AG.Vector3d.ZAxis, vertices[0].DistanceTo(vertices[1]) / 2);
         }
-        return polyline;
       }
+      else
```
clairekuang, line 114: This is a lot of what looks like duplciate conversion code. Can you explain why you decided to do this instead of using the curve converter, as in V2? https://github.com/specklesystems/speckle-sharp/blob/cdf20b4fefb964ed3017bd75eafc351f5cea8df5/Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Other.cs#L310

KatKatKateryna: The tricky part is that we don't actually convert anything to speckle, we are creating an ADB.Curve or ADB.Polyline, so from these loops we can create a native ADB.Region. And then we use the Regions converter for properly assign inner and outer loops, generate mesh etc. Hatches are limited in this functionality.
Also, we don't want to return each segment separately, we need to construct just one loop from the segment collection.

Converters/Autocad/Speckle.Converters.AutocadShared/ToSpeckle/Raw/NurbCurve3dToSpeckleRawConverter.cs
```diff
@@ -0,0 +1,136 @@
...
+    ITypedConverter<ADB.Polyline, SOG.Autocad.AutocadPolycurve> polylineConverter,
+    IConverterSettingsStore<AutocadConversionSettings> settingsStore
+  )
+  {
+    _intervalConverter = intervalConverter;
+    _polylineConverter = polylineConverter;
+    _settingsStore = settingsStore;
+  }
+
+  public SOG.Curve Convert(AG.NurbCurve3d target)
```
clairekuang, line 23: Can you also explain why this conversion is a lot more complex than the v2 nurbscurve3d conversion? 
https://github.com/specklesystems/speckle-sharp/blob/cdf20b4fefb964ed3017bd75eafc351f5cea8df5/Objects/Converters/ConverterAutocadCivil/ConverterAutocadCivilShared/ConverterAutocadCivil.Geometry.cs#L1062

Also, you can use the GetSamplePoints() method on the nurbscurve for the displayvalue. Setting it right now to 100 is very arbitrary, since the curve can be any length. I would set default resolution to be some ratio with the doc units, (eg if doc is in mm, and curve is 1m, the resolution should probably be mm = 1000 segments), with some minimum value.

Converters/Autocad/Speckle.Converters.AutocadShared/ToSpeckle/Geometry/HatchToSpeckleConverter.cs
clairekuang: Why do you need to convert it to an autocad region first, instead of converting it to a speckle region directly?

KatKatKateryna: Regions are much more straightforward to convert:
- They have very clearly typed outed and inner loops
- 1 region guarantees that it's a 1 continuous region (1 hatch can have multiple disconnected parts)
- Regions have a method to extract Brep for the displayValue

Converters/Autocad/Speckle.Converters.AutocadShared/ToSpeckle/Raw/NurbCurve3dToSpeckleRawConverter.cs
KatKatKateryna: as in the comment in the line 25, this is mostly copy-paste from v3 Spline converter, adjusted to the property and method names of NurbCurve3d

clairekuang: Yes, can you explain why you decided to copy the spline converter instead of using the nurbscurve3d conversion from v2, which is a lot less complex?

KatKatKateryna: No reason, it worked perfectly so I used it. Can revert to v2 logic 🙌

Converters/Autocad/Speckle.Converters.AutocadShared/ToSpeckle/Geometry/HatchToSpeckleConverter.cs
clairekuang: This is a commit from v2, with a much simpler hatch conversion that doesn't rely on converting hatches to autocad regions first. The main thing missing is circle and ellipse support: https://latest.speckle.systems/projects/3f895e614f/models/3dc4d7f58f

I would prefer not to add a lot more LOC just to convert to autocad region first, when we can be converting to speckle region directly

KatKatKateryna:
@clairekuang minimized the code and added more test cases to the "objects" file: https://latest.speckle.systems/projects/2295cb26a0/models/7c9dc1dcd8@29b4ad97a5

KatKatKateryna:
These samples broke on receive, because of AutoCAD ToHost Single>Multi conversion, pausing the PR

Converters/Autocad/Speckle.Converters.AutocadShared/ToSpeckle/Geometry/HatchToSpeckleConverter.cs
```diff
@@ -112,8 +105,93 @@ private ADB.Curve PolylineFromLoop(ADB.HatchLoop loop)
...
+        throw new ConversionException($"Hatch loop doesn't contain any segments.");
+      }
+
+      if (loop.Curves.Count > 1) // handle the multi-segment case only with Line segments (not able to produce a Hatch Loop with multiple segments of other types)
+      {
+        foreach (var lineSegment in loop.Curves)
+        {
+          // for each segment, skip the last point: it will be added as a start Point of the next segment. Otherwise, they will overlap and make the curve invalid
+          count = lineSegment is AG.LineSegment2d line
+            ? TryAddPointToPolyline(vertices, polyline, line.StartPoint, 0, count)
```
clairekuang, line 122: Is it necessary to convert linesegment2d into polyline? Does it not work when using Line directly?

Converters/Autocad/Speckle.Converters.AutocadShared/ToSpeckle/Geometry/HatchToSpeckleConverter.cs
```diff
@@ -78,25 +78,19 @@ public SOG.Region Convert(ADB.Hatch target)
 
   private ADB.Curve PolylineFromLoop(ADB.HatchLoop loop)
```
clairekuang, line 79: Can you add a comment at the top of the class, indicating that `Hatch` are complex classes with many edge cases, and the most robust conversion method is converting them to a `Region` first.

clairekuang: Similarly, can you explain that the `Region.CreateFromCurves()` method on line 33 expects database-resident objects, and therefore requires conversion of hatch loops into database entities.

Converters/Autocad/Speckle.Converters.AutocadShared/ToSpeckle/Geometry/HatchToSpeckleConverter.cs
```diff
@@ -112,8 +106,93 @@ private ADB.Curve PolylineFromLoop(ADB.HatchLoop loop)
...
+
+      // if .Polyline is null, read from .Curves
+      if (loop.Curves.Count == 0)
+      {
+        throw new ConversionException($"Hatch loop doesn't contain any segments.");
+      }
+
+      if (loop.Curves.Count > 1) // handle the multi-segment case only with Line segments (not able to produce a Hatch Loop with multiple segments of other types)
+      {
+        foreach (var lineSegment in loop.Curves)
```
clairekuang, line 118: this should be converted to Line entities right? why do you need to convert it to Polyline?

Converters/Autocad/Speckle.Converters.AutocadShared/ToSpeckle/Geometry/HatchToSpeckleConverter.cs
clairekuang, line 79: Also this is a misleading method: should probably call it `ConvertHatchLoopToCurveEntity`. The reason for needing to convert hatch loops to curve entities should be explained as in the above comments

clairekuang:
@bimgeek this is ready for you to test: code is cleaned up, and I fixed a few bugs I found. See https://github.com/specklesystems/speckle-sharp-connectors/pull/856 for more context

clairekuang (approved):

PR #825, opened 2025-05-12
Kateryna/cnx 1687 modifying layer material properties doesnt trigger object

KatKatKateryna:
## Description
Until now, we only triggered Sender model card expiration, if the material directly on the object was changed. Also, it triggered expiration of the object, which was not necessary (object itself has not changed, and materials would be converted again anyway). 

This PR separates and reuses the method for getting all childObjects of a certain layer and all it's sublayers. This method is then used in LayerTableEvent (when parent layer material, color or name changed), and in RenderMaterialAssignment event (in case where material has been changed for the layer). The IDs of the objects returned from this method are written into the list _ChangedObjectIdsInGroupsOrLayers_ , aimed to trigger the Model card expiration, but not expiration of the objects themselves. 
:
image
As in the screenshot, assuming we sent something from the layer Default (or the entire layer), the Model Card expiration will be triggered if the material changes on the object, on the layer "Default", and on the layer "Layer02". At the same time, the sent object will not be expired and will not need to be converted again. 

## Checklist:


clairekuang (changes requested):
This is fine for objects with material that is assigned by layer, but what about objects with material assigned by object? in this case, the expiration for the layer with changed material should change, but we don't want to expire the object too I think

clairekuang:
Are we not storing layer ids? I would just have expiration checks for layers (which would include color changes too), instead of cheacking layers for every object...this seems pretty inefficient

KatKatKateryna:
@clairekuang revised after our chat to not trigger object expiration

oguzhankoral:
> Are we not storing layer ids? I would just have expiration checks for layers (which would include color changes too), instead of cheacking layers for every object...this seems pretty inefficient

@clairekuang object tracking and caching on atomic object level. I believe the approach is correct - we add that objects into `ChangedObjectIdsInGroupsOrLayers` because we won't reconvert object we will just change its proxy structure.

PR #833, opened 2025-05-14
Feat(gh): add tracking for metrics

oguzhankoral:
Worked on top of @adamhathcock's branch and changed things according to needs.

We pass account id through Selector component to get account properly instead always relying on server url

pass_account_id

## HOW TO TEST
disable the code in `TrackEvent` function
image


Connectors/Rhino/Speckle.Connectors.GrasshopperShared/HostApp/SpeckleResource.cs
```diff
@@ -6,15 +6,15 @@
 
 namespace Speckle.Connectors.GrasshopperShared.HostApp;
 
-public abstract record SpeckleUrlModelResource(string Server, string ProjectId)
+public abstract record SpeckleUrlModelResource(string? AccountId, string Server, string ProjectId)
```
clairekuang, line 9: Add comment here, noting that if the user inputs a model url string, this will not contain account info (and that's why the accountID is nullable in the record resource)

clairekuang (changes requested):
Keep in mind we would probably want to refactor analytics tracking into its own tracker class, especially if we want to add node creation/node run tracking in the future.

For now, just add POC inline comments to the tracking events so we can find these more easily if we refactor.

clairekuang:
Tested: check below for metrics from ghv2 and this pr:
image
Items in yellow are missing from current receive metrics:
-"auto" prop, indicating auto-receive and auto-send setting (not super important)
- "sourceHostAppVersion" important
- "workspace_id" prbly important
- "isMultiplayer" (nice, but not a blocker)
Core version looks wrong, shouldn't this be our sdk version?

oguzhankoral:
> Tested: check below for metrics from ghv2 and this pr: image Items in yellow are missing from current receive metrics:
> 
> * "auto" prop, indicating auto-receive and auto-send setting (not super important)
> * "sourceHostAppVersion" important
> * "workspace_id" prbly important
> * "isMultiplayer" (nice, but not a blocker)
> 
> Item in red: source host app is reporting "Grasshopper 8" instead of "grasshopper" Core version looks wrong, shouldn't this be our sdk version?

- `auto`: will do
- `sourceHostAppVersion`: is not something we track with next-gen, even we dont have that info anywhere on receive?
- `workspace_id`: we have to abuse `SpeckleUrlModelResource` with nullable workspace id which is fine, will do
- `isMultiplayer`: will do

clairekuang (approved):

PR #834, opened 2025-05-14
AccountService and AccountManager should use interfaces

adamhathcock:
RootObjectUnpacker is done by AddConnectors

This matches the rest of the codebase.


bjoernsteinhagen (approved):
Looks legit to me - @adamhathcock any reason why we're holding off on this?

adamhathcock:
Covers this too https://github.com/specklesystems/speckle-sharp-connectors/pull/839

PR #836, opened 2025-05-14
Kateryna/cnx 1498 rhino mesh conversion problem on some geometries

KatKatKateryna:
## Description
The issue was reported on the forum showed that Breps that are very far away from origin and have small dimensions (compared to the document tolerance), are sent "broken" to Speckle. This revealed 3 separate issues:
- Rhino API commands extracting meshes from Brep, Extrusion, Hatch and SubD returned inaccurate results when the object is far from origin (1e6 units and more)
- Rhino API commands extracting meshes from Brep, Hatch and SubD returned inaccurate results for small objects, given our default value for MeshingParameters  minimunEdgeLength=0.05
- Speckle MeshToSpeckleConverter returns inaccurate results when the object is far from origin (1e6 units and more). Mesh converter is used in other converters mentioned above (brep, extrusion, hatch, subd), so it was adding even more inaccuracies after e.g. Brep Mesh was extracted. 
 

## User Value
Being able to send objects containing display meshes of any side, from any location in Rhino model. 

## Changes:
1. Making sure that MeshingParameters  are not too large for the given geometry: solved by introducing 'GetAccurateMinEdgeLegth' method. It will only be called if 'highPrecision'=true for DisplayMeshExtractor.GetGeometryDisplayMesh method. Also, it will only affect Breps by identifying any very short edges and adjusting MeshingParameters accordingly.
2. Making sure that the geometry is moved to origin for all distance-sensitive operations: extracting meshes from Rhino geometry, and converting meshes to Speckle. I introduced 2 level of checks for this:
    - adding bool ModelFarFromOrigin, which is a very simple check calculated once in the beginning of Send operation, and will let 99% of Rhino models skip translations back and forth from origin, and convert as usual. 
    - adding method 'TryGetTranslationVector' which will check if the geometry itself is far from origin. If so, the duplicate of geometry will be moved to origin for mesh extraction and mesh conversion.
    - If the previous method returned valid Vector, use if after all operations to move converted Speckle geometry back to the original object location. 

The changes will affect Mesh, Brep, Extrusion, Hatch and SubD converters. Every change ensures that geometry will only be affected by it if really needed (if geometry would have been sent broken otherwise). Most of models and geometries will not be affected and will be converted as usual. 

The inefficiency that I can see is duplicating 3-step logic for all affected converters (move, convert mesh, move back). But this is hard to avoid, because otherwise we might need to introduce a new dependency with injected _meshConverter. 

## Screenshots:
User reported model before:
image
image
After:
image

Rhino performance testing model (scaled down and moved far away) screenshots are attached in Linear.

Rhino object model moved far away from origin (before):
image
image
image

After:
image
image
image

## Validation of changes:

Benchmark Send operations were tested with Rhino performance model and the reported handrail model. There seems to be no visible performance implications. Details in Linear. 

## Checklist:


clairekuang (changes requested):
Some high-level comments:
- Let's add some documentation reference, I looked into it and the source is that rhino operations on geometry is double precision, but mesh is only single precision: https://wiki.mcneel.com/rhino/farfromorigin .
- The moving geometry decision I get, but let's keep the logic to the Mesh converter and displaymeshextractor: 1)there's no need to store the bool in the conversion settings (esp if geometry far away isn't selected by user) only to recalculate later per object 2) avoid copy paste code in the other converters

Converters/Rhino/Speckle.Converters.RhinoShared/RhinoConversionSettingsFactory.cs
```diff
@@ -13,5 +13,18 @@ IConverterSettingsStore<RhinoConversionSettings> settingsStore
   public RhinoConversionSettings Current => settingsStore.Current;
 
   public RhinoConversionSettings Create(RhinoDoc document) =>
-    new(document, unitsConverter.ConvertOrThrow(RhinoDoc.ActiveDoc.ModelUnitSystem));
+    new(document, unitsConverter.ConvertOrThrow(RhinoDoc.ActiveDoc.ModelUnitSystem), ModelFarFromOrigin());
+
+  /// <summary>
+  /// Quick check whether any of the objects in the scene might be located too far from origin and cause precision issues during meshing.
+  /// </summary>
+  private bool ModelFarFromOrigin()
```
clairekuang, line 21: I would remove this method, and create an extension method on GeometryBase instead to check for distance > 1e6 and return transform if true

Converters/Rhino/Speckle.Converters.RhinoShared/ToSpeckle/Meshing/DisplayMeshExtractor.cs
```diff
@@ -43,16 +44,21 @@ public static RG.Mesh GetDisplayMesh(RhinoObject obj)
     return joinedMesh;
   }
 
-  public static RG.Mesh? GetGeometryDisplayMesh(RG.GeometryBase geometry)
+  /// <summary>
+  /// Extracting Rhino Mesh from Rhino GeometryBase using specified MeshingParameters settings, e.g. minimumEdgeLength.
+  /// </summary>
+  public static RG.Mesh GetGeometryDisplayMesh(RG.GeometryBase geometry, bool highPrecision = false)
```
clairekuang, line 50: this should be a simple check: 1) check geometry distance from origin with extension method 2) if too large, move to origin with extension method transform, convert, move back

Converters/Rhino/Speckle.Converters.RhinoShared/ToSpeckle/Meshing/DisplayMeshExtractor.cs
```diff
@@ -76,4 +82,77 @@ public static RG.Mesh GetDisplayMesh(RhinoObject obj)
     joinedMesh.Append(renderMeshes);
     return joinedMesh;
   }
+
+  /// <summary>
+  /// Calculating optimal meshing parameter 'minimumEdgeLength' for the given geometry.
+  /// </summary>
+  private static double GetAccurateMinEdgeLegth(RG.GeometryBase geometry)
```
clairekuang, line 89: the min edge length seems like a separate issue, correct?

Converters/Rhino/Speckle.Converters.RhinoShared/ToSpeckle/Meshing/DisplayMeshExtractor.cs
```diff
@@ -76,4 +82,77 @@ public static RG.Mesh GetDisplayMesh(RhinoObject obj)
...
+  public static RG.Mesh MoveToOriginAndGetDisplayMesh(
+    RG.GeometryBase geometry,
+    bool modelFarFromOrigin,
+    out RG.Vector3d? vectorToOriginalGeometry
+  )
+  {
+    vectorToOriginalGeometry = null;
+
+    // 1. General check: if Model is NOT far from origin (99% of Rhino models): extract meshes as usual
+    if (!modelFarFromOrigin)
```
clairekuang, line 113: change these redundant checks, just check for specific geometry being far from origin

Converters/Rhino/Speckle.Converters.RhinoShared/ToSpeckle/Meshing/DisplayMeshExtractor.cs
```diff
@@ -76,4 +82,77 @@ public static RG.Mesh GetDisplayMesh(RhinoObject obj)
...
+    return geometryToMesh is RG.Mesh movedMesh ? movedMesh : GetGeometryDisplayMesh(geometryToMesh, true);
+  }
+
+  /// <summary>
+  /// Getting translation vector from origin to the Geometry bbox Center (if geometry is far from origin and translation needed)
+  /// </summary>
+  /// <returns>
+  /// True and the vector from origin to Geometry bbox center (if translation needed), otherwise false and zero-length vector.
+  /// </returns>
+  private static bool TryGetTranslationVector(RG.GeometryBase geom, out RG.Vector3d vector)
```
clairekuang, line 136: This is misleading, tryget implies there's some failure to retrieve translation vector, it does not indicate not needing to retrieve translation vector. In any case, I would just have this be the output of the distance check extension method, like so:
```
public bool IsFarFromOrigin(this GeometryBase geo, out Vector translationVector) { 
// test for distance and set bool
// set translation vector default to 0, calculate vector if bool = true
}

File: Converters/Rhino/Speckle.Converters.RhinoShared/ToSpeckle/Raw/BrepToSpeckleConverter.cs [hunk 6]
```diff
@@ -28,8 +28,19 @@ public SOG.BrepX Convert(RG.Brep target)
   {
     var brepEncoding = RawEncodingCreator.Encode(target, _settingsStore.Current.Document);
 
-    var displayMesh = DisplayMeshExtractor.GetGeometryDisplayMesh(target);
-    List<SOG.Mesh> displayValue = displayMesh is null ? new() : new() { _meshConverter.Convert(displayMesh) };
+    // Extracting and converting meshes
+    // 1. If needed, move geometry to origin before all operations; extract Rhino Mesh
```
clairekuang, line 32: remove this code, should be handled in displaymesh extractor

Converters/Rhino/Speckle.Converters.RhinoShared/ToSpeckle/Raw/ExtrusionToSpeckleConverter.cs
```diff
@@ -28,8 +28,19 @@ public SOG.ExtrusionX Convert(RG.Extrusion target)
   {
     var extrusionEncoding = RawEncodingCreator.Encode(target, _settingsStore.Current.Document);
 
-    var displayMesh = DisplayMeshExtractor.GetGeometryDisplayMesh(target);
-    List<SOG.Mesh> displayValue = displayMesh is null ? new() : new() { _meshConverter.Convert(displayMesh) };
+    // Extracting and converting meshes
+    // 1. If needed, move geometry to origin before all operations; extract Rhino Mesh
```
clairekuang, line 32: remove this code, should be handled in displaymesh extractor

Converters/Rhino/Speckle.Converters.RhinoShared/ToSpeckle/Raw/HatchToSpeckleConverter.cs
```diff
@@ -38,8 +38,20 @@ public SOG.Region Convert(RG.Hatch target)
 
     // create display mesh from region by converting to brep first
     var brep = RG.Brep.TryConvertBrep(target);
-    var displayMesh = DisplayMeshExtractor.GetGeometryDisplayMesh(brep);
-    List<SOG.Mesh> displayValue = displayMesh is null ? new() : new() { _meshConverter.Convert(displayMesh) };
+
+    // Extracting and converting meshes
```
clairekuang, line 42: remove this code, should be handled in displaymesh extractor

Converters/Rhino/Speckle.Converters.RhinoShared/ToSpeckle/Raw/MeshToSpeckleConverter.cs
```diff
@@ -31,6 +32,25 @@ public SOG.Mesh Convert(RG.Mesh target)
     {
       throw new ValidationException("Cannot convert a mesh with 0 vertices/faces");
     }
+
+    // 1. If needed, move geometry to origin before conversion
```
clairekuang, line 36: Use same logic as in display mesh extractor

Converters/Rhino/Speckle.Converters.RhinoShared/ToSpeckle/Raw/SubDToSpeckleConverter.cs
```diff
@@ -28,8 +28,19 @@ public SOG.SubDX Convert(RG.SubD target)
   {
     var subdEncoding = RawEncodingCreator.Encode(target, _settingsStore.Current.Document);
 
-    var displayMesh = DisplayMeshExtractor.GetGeometryDisplayMesh(target);
-    List<SOG.Mesh> displayValue = displayMesh is null ? new() : new() { _meshConverter.Convert(displayMesh) };
+    // Extracting and converting meshes
+    // 1. If needed, move geometry to origin before all operations; extract Rhino Mesh
```
clairekuang, line 32: remove this code, should be handled in displaymesh extractor

clairekuang (changes requested):

Converters/Rhino/Speckle.Converters.RhinoShared/RhinoConversionSettingsFactory.cs
KatKatKateryna: Removing this check will force the calculation of a Bbox on every object in any Rhino model. If we keep this check, most of Rhino models will avoid individual objects check.

Converters/Rhino/Speckle.Converters.RhinoShared/ToSpeckle/Meshing/DisplayMeshExtractor.cs
KatKatKateryna: my only reason not to have done it already, was the way to inject the mesh converter here. It's a static class, should we just import it as MeshToSpeckle converter, instead of a usual injection of TypedConverter?

Converters/Rhino/Speckle.Converters.RhinoShared/ToSpeckle/Meshing/DisplayMeshExtractor.cs
clairekuang: I think we concluded this was not necessary right? Don't forget to remove if so

Converters/Rhino/Speckle.Converters.RhinoShared/Extensions/GeometryBaseExtensions.cs
```diff
@@ -0,0 +1,23 @@
+namespace Speckle.Converters.Rhino.Extensions;
+
+public static class GeometryBaseExtensions
+{
+  /// <summary>
+  /// Getting translation vector from origin to the Geometry bbox Center (if geometry is far from origin and translation needed)
+  /// </summary>
+  /// <returns>
+  /// Vector from origin to Geometry bbox center (if translation needed), otherwise zero-length vector.
+  /// </returns>
```
clairekuang, line 10: Oh forgot, can you add this in-code comment link to documentation: https://wiki.mcneel.com/rhino/farfromorigin

clairekuang: (and explain this issue is caused by rhino api geometry operations being double-precision, but rhino mesh is single-precision)

clairekuang (approved):

PR #839, opened 2025-05-15
Fix DI usage for MixPanelManager

adamhathcock:
Don't parse slug, just grab it directly.
Use DI interfaces for future testability as they exist already.

Relates to: https://github.com/specklesystems/speckle-sharp-connectors/pull/834


JR-Morgan (changes requested):
Discussed in person. The slug changes are not correct, we want to track the received version source app, not the current host app slug.

Yes, this is nasty, clearly a hack that we're mapping the host app name -> slug.. Totally agree but for now I think it's minor.

The interface changes are good, so if you wanted to bring these out into a seperate PR, or reuse this one, that's good with me

PR #840, opened 2025-05-15
Fix(build): Fix Deep Clean

JR-Morgan:
@adamhathcock It seems that deep-clean has not been working for a while.

```
C:\Program Files\dotnet\sdk\8.0.409\Microsoft.Common.CurrentVersion.targets(5321,5): error MSB3027:
  Could not copy "C:\Speckle Repos\speckle-sharp-connectors\Build\obj\Debug\net8.0\apphost.exe" to "bin\Debug\net8.0\Build.exe".
  Exceeded retry count of 10.
  Failed.
  The file is locked by: "build (25904)" [C:\Speckle Repos\speckle-sharp-connectors\Build\Build.csproj]

C:\Program Files\dotnet\sdk\8.0.409\Microsoft.Common.CurrentVersion.targets(5321,5): error MSB3021:
  Unable to copy file "C:\Speckle Repos\speckle-sharp-connectors\Build\obj\Debug\net8.0\apphost.exe" to "bin\Debug\net8.0\Build.exe".
  The process cannot access the file 'C:\Speckle Repos\speckle-sharp-connectors\Build\bin\Debug\net8.0\Build.exe' because it is being used by another process. [C:\Speckle Repos\speckle-sharp-connectors\Build\Build.csproj]

```

I've investigated, and it looks like `deep-clean` attempts to build the entire solution. But because we are running `Build.csproj` which is in the solution, some of the files are locked by the build process. Thus the build process cannot rebuild it's self, and the whole deep-clean fails.

---

In this PR, I've made deep-clean just clean files + restore. It doesn't attempt to build. I figured this is more useful that the current state.

### DIFF
--- a/Build/Program.cs
+++ b/Build/Program.cs
@@ -28,14 +28,7 @@ if (args.Length > 1)
   args = new[] { arguments.First() };
   //arguments = arguments.Skip(1).ToList();
 }*/
-void Build(string solution, string configuration)
-{
-  Console.WriteLine();
-  Console.WriteLine();
-  Console.WriteLine($"Building solution '{solution}' as '{configuration}'");
-  Console.WriteLine();
-  Run("dotnet", $"build \".\\{solution}\" --configuration {configuration} --no-restore");
-}
+
 void Restore(string solution)
 {
   Console.WriteLine();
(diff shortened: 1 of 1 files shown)


adamhathcock (approved):
Remove build part of deep clean

JR-Morgan:
For posterity, there was a little confusion on this subject from my end.

deep-clean was intended to error out in the this way. because it still technically works, cleans and rebuilds everything that matters, but the job reports itself as failed.

I think it's valid to have a clean that builds too, but I think it's more useful if it just cleans (as maybe you're cleaning to install connectors via installers, or wanting to swap build configurations....

PR #842, opened 2025-05-16
chore(grasshopper): updates sample files to rhino 7 only components

clairekuang:
The current sample files contained a component that was only available in rhino 8, showing an error when opened in rhino 7

This updates the sample files to only use rhino 7 components so they can be used in both versions


JR-Morgan (approved):

JR-Morgan (changes requested):
I think these will need to be moved into the root of the sample files folder for csproj to copy, and installers to pickup up.

clairekuang:
> I think these will need to be moved into the root of the sample files folder for csproj to copy, and installers to pickup up.

is it not already?

JR-Morgan (approved):

JR-Morgan:
Ah yes they are 🤦‍♂️ dunno what I was thinking

PR #843, opened 2025-05-16
fix (Revit) :Hermite Spline conversion when NURBS conversion failing

kekesidavid:
An exception was thrown by NurbSpline.CreateCurve
because Revit is stricter than Rhino regarding the input parameters for NURBS curves.
Exception message:
"The multiplicities of other interior knots must be at most degree - 2."
The solution falls back to using displayValue.

### DIFF
--- a/Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/CurveConverterToHost.cs
+++ b/Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/CurveConverterToHost.cs
@@ -46,6 +46,7 @@ public class CurveConverterToHost : ITypedConverter<SOG.Curve, DB.Curve>
       }
       // An exception was thrown by NurbSpline.CreateCurve
       // because Revit is stricter than Rhino regarding the input parameters for NURBS curves.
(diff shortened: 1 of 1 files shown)


Converters/Revit/Speckle.Converters.RevitShared/ToHost/Raw/Geometry/CurveConverterToHost.cs
```diff
@@ -33,9 +38,34 @@ public DB.Curve Convert(SOG.Curve target)
...
+      try
+      {
+        //var knots = speckleKnots.GetRange(0, pts.Count + speckleCurve.degree + 1);
+        var curve = DB.NurbSpline.CreateCurve(target.degree, speckleKnots, pts, weights);
+        return curve;
+      }
+      // An exception was thrown by NurbSpline.CreateCurve
+      // because Revit is stricter than Rhino regarding the input parameters for NURBS curves.
+      // Exception message:
+      // "The multiplicities of other interior knots must be at most degree - 2."
```
clairekuang, line 50: I would add a comment to say that this happens on semi-circle curves that were converted to nurbs

clairekuang (approved):

PR #844, opened 2025-05-16
feat(civil 3d): adds pressure pipe base curve, display value, and partdata

clairekuang:
## Description

Adds pressure pipe base curve, displayvalue, and partdata.
This was necessary to handle separately because they do not inherit from `Part` as with regular pipes, they inherit from `PressurePart` and therefore need their own case in the basecurveextractor, displayvalueextractor and propertiesextractor

- Note: for basecurve extraction, these methods are only available in Civil3d 2024+
## User Value

Now you get your pressure pipes with mesh and props.
Community post: https://speckle.community/t/civil-3d-pressure-pipes-not-exported/17841

## Changes:

civil converter

## Screenshots:

- After: https://latest.speckle.systems/projects/3f895e614f/models/1737754fb4

## Validation of changes:

Sent test file

### DIFF
--- a/Converters/Civil3d/Speckle.Converters.Civil3dShared/ToSpeckle/Properties/PartDataExtractor.cs
+++ b/Converters/Civil3d/Speckle.Converters.Civil3dShared/ToSpeckle/Properties/PartDataExtractor.cs
@@ -16,12 +16,31 @@ public class PartDataExtractor
   /// <returns></returns>
   public Dictionary<string, object?>? GetPartData(CDB.Entity entity)
   {
-    if (entity is CDB.Part part)
+    return entity switch
     {
-      return ParsePartData(part.PartData);
+      CDB.Part part => ParsePartData(part.PartData),
+      CDB.PressurePart pressurePart => ParsePartData(pressurePart.PartData),
+      _ => null
+    };
+  }
+
+  private Dictionary<string, object?> ParsePartData(CDB.PressureNetworkPartData partData)
+  {
+    var result = new Dictionary<string, object?>();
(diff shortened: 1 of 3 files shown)


Converters/Civil3d/Speckle.Converters.Civil3dShared/ToSpeckle/Properties/PartDataExtractor.cs
```diff
@@ -16,12 +16,31 @@ public PartDataExtractor() { }
...
+    foreach (CDB.PressurePartProperty prop in partData)
+    {
+      if (!prop.HasValue)
+      {
+        continue; // don't send null props
+      }
+
+      if (!result.ContainsKey(prop.DisplayName))
+      {
+        result.Add(prop.DisplayName, prop.Value);
```
bjoernsteinhagen, line 39: for `CDB.Part`, we create a complex dictionary with value, name, context, and units etc. For `CDB.PressurePart`, we simply store the value directly without "metadata" .. intentional? if so, ignore this :)

clairekuang: Intentional, apparently pressurepart data is a lot simpler and doesn't contain units or context eg

bjoernsteinhagen: Gotcha. I'm happy!

bjoernsteinhagen (approved):

PR #853, opened 2025-05-20
feat(grasshopper): deconstruct and create empty props

clairekuang:
## Description

Adds an option to handle Speckle Properties as an input for `Deconstruct`, which exposes all dictionary entries
Adds a right-click menu item to `Create Properties` node to allow users to create empty properties
## User Value

Better property UX, to retrieve values quickly

## Changes:

-Deconstruct node
-create properties node

## Screenshots:

{7A8704E8-1B40-48DC-BA1E-9E94E493E2D1}

## Validation of changes:

Simple testing of both nodes.
Haven't tested file save behavior.
Undo currently does not undo "create empty Properties" actions.

## Checklist:


bjoernsteinhagen (approved):
tested and looks good to me! 😃

PR #857, opened 2025-05-22
fix(revit): workflow sketchup to revit object material assignment and colors

bjoernsteinhagen:
## Description

In [CNX-1707](https://linear.app/speckle/issue/CNX-1707/revit-curves-and-meshes-in-blocks-come-as-duplicated), we introduced a modification to the `applicationId` to solve the instancing issue. However, this broke the mapping of `applicationId` in the proxies to the actual objects. For example, Sketchup to Revit workflow was broken: materials were being created in Revit but not assigned to the object(s). 

We will reconsolidate and do this properly:
- [CNX-1797 Revit - Refactor Revit Material Handling System](https://linear.app/speckle/issue/CNX-1797/revit-refactor-revit-material-handling-system)
- [CNX-1769 Revit - Refactoring BakeObjects](https://linear.app/speckle/issue/CNX-1769/revit-refactoring-bakeobjects)

## User Value

Material assignment on load.

## Changes:

- Moved modification of `applicationId` to before `BakeMaterials`
- Have the `revitToHostCacheSingleton.MaterialsByObjectId` use the modified `applicationId` and not the original (that way, we keep the [CNX-1707](https://linear.app/speckle/issue/CNX-1707/revit-curves-and-meshes-in-blocks-come-as-duplicated) fix).

## Validation of changes:

Tested load for: **Sketchup to Revit**, **Rhino to Revit** and **Revit to Revit**

image

## Checklist:

### DIFF
--- a/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Receive/RevitHostObjectBuilder.cs
+++ b/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Receive/RevitHostObjectBuilder.cs
@@ -101,7 +101,68 @@ public sealed class RevitHostObjectBuilder(
       unpackedRoot.ObjectsToConvert.ToList()
     );
 
-    // 2 - Bake materials
+    // NOTE: below is 💩... https://github.com/specklesystems/speckle-sharp-connectors/pull/813 broke sketchup to revit workflow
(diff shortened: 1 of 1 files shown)


oguzhankoral (approved):
Tentaively approving. This mess already became a mountain. Need to untangle before too late. We need to also extract logics outside of host object builder.

PR #862, opened 2025-05-22
fix: passes correct arguments to ui on send results

didimitrie:
image

It would be great if we can release again after this merge, as it's blocking [CNX-598: Version message](https://linear.app/speckle/issue/CNX-598/version-message)


oguzhankoral (approved):

adamhathcock:
@oguzhankoral @didimitrie @JR-Morgan hotfix to main?

PR #867, opened 2025-05-25
feat(grasshopper): add inherit names functionality to variable parameter components

bjoernsteinhagen:
## Description
Adds inherit names functionality to `CreateCollection` and `CreateProperties` components. 

Solves [CNX-1820](https://linear.app/speckle/issue/CNX-1820/inherit-names-for-create-collection-and-create-properties-nodes).

## User Value
Quickly renaming of input parameters to match connected sources instead of manually typing names.

#### Two Ways to Inherit Names

**Method 1: MANUAL (Tab Key)**
- User holds TAB → connects wire → parameter inherits name once

**Method 2: AUTOMATIC (Always Inherit)**  
- User enables setting → ALL parameters inherit names automatically → Names update when source names change
- When enabled, renaming is greyed out
- Resizing also works with name updates

#### New Classes / Modified Classes

`VariableParameterComponentBase` - SHARED BASE CLASS
- Handles "Always inherit names" setting
- Creates SpeckleVariableParam instances

`SpeckleVariableParam` - INDIVIDUAL PARAMETERS
- Detects Tab key presses
- Manages own event subscriptions
- Inherits names from connected sources 
- Updates automatically when sources change 

Other changes:
- `KeyWatcher` - utility to track Tab key state during parameter connections
- Updated `CreateCollection` and  `CreateSpeckleProperties` accordingly
- Added banner to indicate when `AlwaysInheritNames == true`

## Validation of changes:

https://github.com/user-attachments/assets/bddc6add-e018-4095-bd16-303f92270cdf

## Checklist:


bjoernsteinhagen:
> > @bjoernsteinhagen probably should merge #868 in before this pr
> 
> Testing and reviewing #868 now!

With #868 now in, I resolved conflicts, retested on my end and all still behaving as before (with the changes of #868 reflecting)

bjoernsteinhagen:
Always inherit functionality included, PR description and video valdiation updated.

bjoernsteinhagen:
tested again with all gh merged changes from other tickets - all good @clairekuang

clairekuang (changes requested):
Tested, seems like there's some inconsistency of behavior with _individual_ parameter name inheritance setting:
1. Inherit all names is turned off, but a previously inherited name shows the inherit name flag:
image
2. inherit all names is off, and a non-inherited name does not have the flag (expected)
image

I would just remove the extra menu item on individual parameters in all cases. 

Also, while inheriting all names, the component should auto update when an input param name has changed:
{F393FD6D-8059-4007-AE58-B0A411AC87C0}
Currently, you have to re-pipe the input to have the new name reflected

bjoernsteinhagen:
> I would just remove the extra menu item on individual parameters in all cases.

Yeah it's empty because nothing was connected to `Property4`, hence nothing to inherit. 

If I remove the extra menu item, would this apply to the tab-key functionality too, then @clairekuang? This allows for auto-inheritance on an individual parameter basis.

clairekuang:
I like the tab-key feature! That can stay, and work as usual when inherit all names is off

clairekuang (approved):
Looks good! nice work

PR #868, opened 2025-05-25
feat(grasshopper): automatically adds new param for create collection and properties

clairekuang:
## Description
Mimicking the behavior of Grasshopper's native `Merge` component, this pr automatically adds a new input param if a source has been added to the last param on the components:
- Create Collection
- Create Properties
Expected behavior: New input should be created only on the addition of a source to the last input param. If any source is disconnected, this component shouldn't change inputs. Check native gh merge component for same behavior

Note:
- Also fixed an issue with the Create Properties component, to allow for empty props to be created. Previous node also was allowing some inputs like point, when we do not support any non-string/bool/number inputs.

## User Value

Easier UX when creating many inputs for props or collections

## Changes:

- Create Properties
- Create Collection

## To-do before merge:

- [ ] Check for same behavior as merge node

## Screenshots:

{FBD5D75D-1827-4D26-A5B8-A6B61E669678}

## Validation of changes:

Simple testing of adding sources, removing sources, and adding and deleting params


bjoernsteinhagen (approved):
Solid 😎

PR #869, opened 2025-05-26
chore (all sharp) renamed speckle beta to speckle everywhere

kekesidavid:
- Renamed "Speckle (Beta)" to "Speckle" everywhere with a simple find and replace
- Deleted the leftover "Beta" texts
- Should be checked and tested by connector owners


bjoernsteinhagen:
Are we happy with this wording @bimgeek?

image

clairekuang:
Tested Autocad, no conflicts even before 2.23

PR #870, opened 2025-05-26
feat(grasshopper): add direct casting from speckle object to geometry types

bjoernsteinhagen:
## Description

Adds direct casting from speckle objects to other native gh geometry types. Namely:

- `Point`
- `Curve`
- `Line`
- `Brep`
- `Mesh`
- `SubD`
- `Surface`
- `Pointcloud`

Fixes [CNX-1741](https://linear.app/speckle/issue/CNX-1741/add-direct-casting-from-speckle-object-to-geometry-types).

## User Value

Direct casting (stems for a user request). 

## Changes:

`SpeckleObjectWrapper.cs` has the `CastTo` method extended with another conditional. This is the `TryUnifiedCast` which essentially tries to "wrap" all type conversion logic in to a condensed method. But, native gh conversions can become quite verbose. This was seen as the best solution considering the strict typing requirements.

## Validation of changes:

Note that mixed geometry will light up red if we directly pipe into a native specific component (e.g. `Curve` or `Brep`) as opposed to the generic `Geometry` component. This, however, mimics native behaviour and the conversions still do take place. Users just need to filter out the nulls.

image

## Checklist:

### DIFF
--- a/Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Parameters/SpeckleObjectWrapper.cs
+++ b/Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Parameters/SpeckleObjectWrapper.cs
@@ -290,126 +290,110 @@ public partial class SpeckleObjectWrapperGoo : GH_Goo<SpeckleObjectWrapper>, IGH
 
     var targetType = typeof(T);
 
-    switch (targetType)
+    // try our "unified" conversion
+    if (TryUnifiedCast<T>(Value.GeometryBase, out var result))
     {
-      case var _ when targetType == typeof(GH_Point):
-        return TryCastToPoint(ref target);
-
-      case var _ when targetType == typeof(GH_Curve):
-        return TryCastToCurve(ref target);
-
-      case var _ when targetType == typeof(GH_Line):
-        return TryCastToLine(ref target);
-
-      case var _ when targetType == typeof(GH_Brep):
-        return TryCastToBrep(ref target);
-
-      case var _ when targetType == typeof(GH_Mesh):
-        return TryCastToMesh(ref target);
-
-      case var _ when targetType == typeof(GH_SubD):
-        return TryCastToSubD(ref target);
-
-      case var _ when targetType == typeof(GH_Surface):
-        return TryCastToSurface(ref target);
-
-      case var _ when targetType == typeof(GH_PointCloud):
-        return TryCastToPointcloud(ref target);
-
-      case var _ when targetType == typeof(IGH_GeometricGoo):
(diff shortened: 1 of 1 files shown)


Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Parameters/SpeckleObjectWrapper.cs
```diff
@@ -283,17 +283,122 @@ public override bool CastFrom(object source)
...
+      PointCloud? pointCloud = null;
+      if (GH_Convert.ToPointCloud(geometry, ref pointCloud, GH_Conversion.Both))
+      {
+        convertedValue = new GH_PointCloud(pointCloud);
+        success = true;
+      }
+    }
+#endif
+
+    if (success && convertedValue != null)
```
clairekuang, line 393: Also, you don't need to set success above: a null converted value would indicate false, so you can test for this case and return false before setting the result if needed

Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Parameters/SpeckleObjectWrapper.cs
```diff
@@ -283,17 +283,122 @@ public override bool CastFrom(object source)
...
+
+    // try our "unified" conversion
+    if (TryUnifiedCast<T>(Value.GeometryBase, out var result))
     {
-      target = (T)(object)GH_Convert.ToGeometricGoo(Value.GeometryBase);
+      target = result;
       return true;
     }
 
+    // handle IGH_GeometricGoo special case
```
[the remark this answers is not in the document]
bjoernsteinhagen: Done in [refactor: cast to geo first](https://github.com/specklesystems/speckle-sharp-connectors/pull/870/commits/7a4261328d5228ad39289c243363f7134d6f0c0d) :)

Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Parameters/SpeckleObjectWrapper.cs
bjoernsteinhagen: Lol, good point. Don't know why I had that. Done in [chore: remove redundant success](https://github.com/specklesystems/speckle-sharp-connectors/pull/870/commits/08b41b0b429892245958e792955dd5373663a5a7)

bjoernsteinhagen:
@clairekuang we ran into cyclomatic complexity errors on the `TryUnififiedCast` with the addition of the missing types. So, pivoted back to solution of separate methods per cast. This way, our `CastTo` method is at least much cleaner and easier to understand with the switch statements.

clairekuang (changes requested):

PR #881, opened 2025-05-30
Add Tekla 2025 properly to sln

adamhathcock:


adamhathcock:
Probably not needed after https://github.com/specklesystems/speckle-sharp-connectors/pull/879

PR #882, opened 2025-05-30
chore (rhino, autocad) updated converters after text class update

kekesidavid:
Updated converters for Autocad and Rhino after Text class property updates.


Sdk/Speckle.Common.MeshTriangulation/Speckle.Common.MeshTriangulation.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>net48;net6.0;net8.0</TargetFrameworks>
+	<Configurations>Debug;Release;Local</Configurations>
```
clairekuang, line 5: should this change be present?

PR #885, opened 2025-06-02
fix (tekla) replaced svg image for tekla

kekesidavid:
- the svg resource somehow has a resolution (should it?)
- replaced the 556x556 image with a 24x24


bjoernsteinhagen (approved):
Confirmation:

### Before

image

### After

image

PR #887, opened 2025-06-03
fix(logs) Only log errors and observe them if they're speckle related. 

adamhathcock:
Others log if possible and let bomb out the app


DUI3/Speckle.Connectors.DUI/ContainerRegistration.cs
```diff
@@ -40,10 +40,22 @@ public static void UseDUI(this IServiceProvider serviceProvider)
     {
       try
       {
-        serviceProvider
-          .GetRequiredService<ILoggerFactory>()
-          .CreateLogger("UnobservedTaskException")
-          .LogError(args.Exception, "Unobserved task exception");
+        var stackTrace = args.Exception.StackTrace;
+        if (stackTrace?.IndexOf("Speckle", StringComparison.InvariantCultureIgnoreCase) <= 0)
```
JR-Morgan, line 44: In my testing. The `StackTrace` here appears to frequently be `null` on the `AggregateException` that is `args.Exception`
And instead, the inner exception has the real stack trace.

It looks like `.StackTrace` does not include the inner exception stack traces. I think this is important to ensure this logic is reliable.

JR-Morgan (changes requested):

DUI3/Speckle.Connectors.DUI/ContainerRegistration.cs
JR-Morgan: image

adamhathcock: I didn't realize it was always an Aggregate exception.  I'm now handling it as such

DUI3/Speckle.Connectors.DUI/ContainerRegistration.cs
```diff
@@ -53,10 +56,25 @@ public static void UseDUI(this IServiceProvider serviceProvider)
...
-      {
-        args.SetObserved();
-      }
     };
   }
+
+  private static bool TestAndLogStacktrace(IServiceProvider serviceProvider, Exception exception)
+  {
+    var stackTrace = exception.StackTrace;
+    if (stackTrace is null || stackTrace.IndexOf("Speckle", StringComparison.InvariantCultureIgnoreCase) <= 0)
```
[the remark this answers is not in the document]
JR-Morgan: I gave you some bad information 
image

`GetFormattedString` does not do what I thought it did

PR #895, opened 2025-06-04
Introduce receive manager

adamhathcock:
requires https://github.com/specklesystems/speckle-sharp-connectors/pull/892

introduces a manager to handle exceptions similar to: https://github.com/specklesystems/speckle-sharp-connectors/pull/871


JR-Morgan (approved):
Tested Rhino, Revit, Autoacd

PR #903, opened 2025-06-06
fix (Rhino) Missing group from applicationIdMap fix

adamhathcock:
Looks like a debug but fixed anyway as it was spamming

image


Connectors/Rhino/Speckle.Connectors.RhinoShared/HostApp/RhinoGroupBaker.cs
```diff
@@ -46,6 +48,19 @@ string baseLayerName
...
+    string oldObjId,
+    Dictionary<string, IReadOnlyCollection<string>> applicationIdMap
+  )
+  {
+    if (applicationIdMap.TryGetValue(oldObjId, out IReadOnlyCollection<string> value))
+    {
+      return value;
+    }
+
+    return [];
```
JR-Morgan, line 61: We really shouldn't have speckle data with invalid groups where the application ids don't map.

So I would really like to be alerted if we see this as it could indicated a bug on send.

We could replace this `return [];` with a `throw SpeckleException("Group {group} references an application Id {appId} that cannot be mapped");` or similar

JR-Morgan (changes requested):

PR #907, opened 2025-06-10
Don't use static scopes for Grasshopper. Use when necessary

adamhathcock:


clairekuang (approved):
Tested independently with sync and async send and receive nodes
Found doc locking problems when running concurrent send and receive on the same model, but this is probably pre-existing

PR #908, opened 2025-06-10
feat(grasshopper): allows nested properties

clairekuang:
## Description

Allows for properties to be nested. Includes the following changes:
- `Create Properties` can now accept other `PropertyGroup`s as input
- On receive, nested properties will be preserved in their structure instead of flattened
- Get property paths and query properties work as before, using the concatenated keys
- baking should work as before, using the concatenated keys

Also fixes this: https://linear.app/speckle/issue/CNX-1993/getproperties-node-should-return-null-if-prop-doesnt-exist-on-object

## User Value

Create a more organized set of properties for display in the viewer.

## Changes:

- send and receive nodes
- create properties
- casting on propgroup params

## To-do before merge:

- [x] More testing on receive
- [x] Check that inherited name changes on input prop params expire parents to prevent the "wrong" prop keys from being published
- [x] refactor to handle duplicate objectattribute code

## Checklist:


bjoernsteinhagen (approved):
Nice. My notes for things related to this which can be tackled post merge:
- Deconstructing params for `props` of multiple object types should not be allowed / user gets warning
- names of `props` not inheriting well enough / auto-updating / re-computing

PR #911, opened 2025-06-11
Update to SDK 3.4.2

adamhathcock:
Covers https://github.com/specklesystems/speckle-sharp-sdk/releases/tag/3.4.1

3.4.2 is a reversion of the concurrent send fix as it's even more complex than thought.

Database locked errors should be gone for large models that have a single send


JR-Morgan (changes requested):
image

JR-Morgan (approved):
Tested Rhino + Grasshopper

PR #927, opened 2025-06-16
feat(Navisworks): CNX-2009 - Adds saved views filter to Navisworks send operations

jsdbroughton:
Introduces a new filter for selecting saved views during send operations in the Navisworks connector, enabling users to send only visible elements from saved views. Working with saved views to curate data exchanges is the second most common interaction users perform with Navisworks after Saved Sets, which is already implemented.

image

It doesn't, at this time, add the selected viewpoint to Canonical views. I don't think other connectors do either??

image

Only a single saved view can be selected. The `>` separated view names represent the group hierarchy of saved viewpoints.

Additional changes (DRY):
Enhances the build process of hierarchical names for saved items by refactoring code into a reusable helper function, improving maintainability and consistency across filters.

### DIFF
--- a/Connectors/Navisworks/Speckle.Connectors.NavisworksShared/Operations/Send/Filters/NavisworksSavedViewsFilter.cs
+++ b/Connectors/Navisworks/Speckle.Connectors.NavisworksShared/Operations/Send/Filters/NavisworksSavedViewsFilter.cs
@@ -1,0 +1,173 @@
+﻿using Speckle.Connector.Navisworks.Services;
+using Speckle.Connectors.DUI.Exceptions;
+using Speckle.Connectors.DUI.Models.Card.SendFilter;
+using Speckle.Connectors.DUI.Utils;
+
+namespace Speckle.Connector.Navisworks.Operations.Send.Filters;
+
+public class NavisworksSavedViewsFilter : DiscriminatedObject, ISendFilterSelect
+{
+  private readonly IElementSelectionService _selectionService;
+
+  public NavisworksSavedViewsFilter(IElementSelectionService selectionService)
+  {
+    _selectionService = selectionService;
+
+    Items = [];
+    SelectedItems = [];
+
+    GetSavedViews();
+  }
+
+  public string Id { get; set; } = "navisworksSavedViews";
+
+  public string Name { get; set; } = "Saved Views";
+
+  public string Type { get; set; } = "Select";
+
+  public string? Summary { get; set; }
+
+  public bool IsDefault { get; set; }
+
+  public List<string> SelectedObjectIds { get; set; } = [];
+
+  public Dictionary<string, string>? IdMap { get; set; }
+
+  public bool IsMultiSelectable { get; set; }
+
+  public List<SendFilterSelectItem> SelectedItems { get; set; }
+
+  public List<SendFilterSelectItem> Items { get; set; }
+
+  public List<string> RefreshObjectIds()
+  {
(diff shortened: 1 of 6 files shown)


Connectors/Navisworks/Speckle.Connectors.NavisworksShared/Operations/Send/Filters/NavisworksSavedViewsFilter.cs
```diff
@@ -0,0 +1,173 @@
...
+    // THIS IS COMMENTED OUT AS IT IS LEGACY DEFENSIVE BEHAVIOUR - DISCUSSION REQUIRED
+    // if (!savedView.ContainsVisibilityOverrides)
+    // {
+    //   // We check this again as the view settings may have changed in the saved card.
+    //   // If the saved view does not contain visibility overrides, this is effectively everything in the model.
+    //   // This will need to be the documented behaviour.
+    //   throw new SpeckleSendFilterException(
+    //     "Saved view does not contain visibility overrides. This would effectively publish everything in the model."
+    //   );
+    // }
```
jsdbroughton, line 94: Willing to discuss either uncommenting or removing this. An opinion expressed in v2 code was that this was valuable, partly because sends were slower then, but also because this class of viewpoint doesn't hide anything from view; it's no different from sending everything.

This PR keeps the legacy position but commented out.

Connectors/Navisworks/Speckle.Connectors.NavisworksShared/Operations/Send/Filters/NavisworksSavedViewsFilter.cs
```diff
@@ -0,0 +1,173 @@
...
+
+    foreach (NAV.SavedItem item in ((NAV.GroupItem)parentItem).Children)
+    {
+      switch (item.IsGroup)
+      {
+        // THIS IS COMMENTED OUT AS IT IS LEGACY DEFENSIVE BEHAVIOUR - DISCUSSION REQUIRED
+        // case false when item is NAV.SavedViewpoint { ContainsVisibilityOverrides: false }:
+        //   // If the saved view does not contain visibility overrides, this is effectively everything in the model.
+        //   // This will need to be the documented behaviour.
+        //   break;
```
jsdbroughton, line 160: Willing to discuss either uncommenting or removing this. An opinion expressed in v2 code was that this was valuable, partly because sends were slower then, but also because this class of viewpoint doesn't hide anything from view; it's no different from sending everything.

This PR keeps the legacy position but commented out.

Connectors/Navisworks/Speckle.Connectors.NavisworksShared/Operations/Send/Filters/NavisworksSavedViewsFilter.cs
adamhathcock: happy to leave it, no idea if it's correct or not

Connectors/Navisworks/Speckle.Connectors.NavisworksShared/Operations/Send/Filters/NavisworksSavedViewsFilter.cs
adamhathcock: same as before

adamhathcock (approved):
Follows the pre-existing code base well.  Probably needs a round of perf testing as there are intermediate collections being made but otherwise good

PR #933, opened 2025-06-19
fix(grasshopper): object and collection now shown as selected from parent components

clairekuang:
Previously, selecting a component that outputs an object or collection param was not showing the preview of that param as selected.

Adds a case for testing for parent selection on the param.

Also fixes Create Collection to be of output type collectionparam, to enable param previewing


bjoernsteinhagen (approved):
lgtm

PR #935, opened 2025-06-20
(feat) non windows building

adamhathcock:
Thanks to @JR-Morgan I realized that we actually can build on non-windows.  Just need to disable copy steps when not on windows for now.

Also removed mac only solution


oguzhankoral (approved):

JR-Morgan:
@adamhathcock @oguzhankoral I've adjusted the branch protection rules that were previously requring this PR be reviewed by 5 different people.

Now, the requirement for code owners can be bypassed explicitly by repo admins
image

We should use this sparingly and politely. Feel free to use it now, for this PR @adamhathcock

PR #937, opened 2025-06-20
feat (Grasshopper) dev enable account auth by dev token

adamhathcock:
replaces https://github.com/specklesystems/speckle-sharp-connectors/pull/920
Slight refactor to use Account object sooner.
Grasshopper now a new component just for  URL with Token

### DIFF
--- a/Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Components/Dev/TokenUrlComponent.cs
+++ b/Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Components/Dev/TokenUrlComponent.cs
@@ -1,8 +1,12 @@
 using System.Runtime.InteropServices;
 using Grasshopper.Kernel;
+using Speckle.Connectors.Common.Operations;
 using Speckle.Connectors.GrasshopperShared.HostApp;
 using Speckle.Connectors.GrasshopperShared.Parameters;
 using Speckle.Connectors.GrasshopperShared.Properties;
+using Speckle.Connectors.GrasshopperShared.Registration;
+using Speckle.Sdk;
+using Speckle.Sdk.Api;
 
 namespace Speckle.Connectors.GrasshopperShared.Components.Dev;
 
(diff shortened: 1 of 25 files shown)


Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Components/Dev/TokenUrlComponent.cs
```diff
@@ -0,0 +1,133 @@
...
+    // Should notify user if any of this goes wrong.
+
+    var resources = SpeckleResourceBuilder.FromUrlString(input, token);
+    if (resources.Length == 0)
+    {
+      throw new SpeckleException($"Input url string was empty");
+    }
+
+    if (resources.Length > 1)
+    {
```
clairekuang, line 90: similar as above, this case is already handled in `ParseFe2RegexMatch`. this shouldn't ever hit

Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Components/Dev/TokenUrlComponent.cs
```diff
@@ -0,0 +1,133 @@
...
+    var resource = resources.First();
+    using var scope = PriorityLoader.CreateScopeForActiveDocument();
+    var account = resource.Account.GetAccount(scope);
+    if (account != null)
+    {
+      scope.Get<IAccountService>().SetUserSelectedAccountId(account.id);
+    }
+    else
+    {
+      throw new SpeckleException("No account found for server URL");
```
clairekuang, line 103: `GetAccount` can throw a few exceptions, would this block ever hit? we should inform the user if their token is expired or invalid - this doesn't seem to be the most descriptive message

clairekuang (changes requested):
Requested changes mostly around exception handling. Will give it a test after they're addressed

clairekuang:
Added @oguzhankoral as a reviewer for DUI changes ( looked fine to me )

Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Components/Dev/TokenUrlComponent.cs
```diff
@@ -0,0 +1,133 @@
...
+  )
+  {
+    // When input is provided, lock interaction of buttons so only text is shown (no context menu)
+    // Should perform validation, fill in all internal data of the component (project, model, version, account)
+    // Should notify user if any of this goes wrong.
+
+    var resources = SpeckleResourceBuilder.FromUrlString(input, token);
+    if (resources.Length == 0)
+    {
+      throw new SpeckleException($"Input url string was empty");
```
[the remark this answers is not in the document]
adamhathcock: I did what SpeckleOperationWizard does.  Can change to your suggestion though

Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Components/Dev/TokenUrlComponent.cs
adamhathcock: ParseFe2RegexMatch allows for more than one value because of addtionalModels?  Just doing what the other callers do.  Maybe we should only ever allow one?

Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Components/Dev/TokenUrlComponent.cs
adamhathcock: It's a new get account that basically always returns empty after checking fallbacks and whatnot.  Should not return a null ever?

DUI3/Speckle.Connectors.DUI/Models/Card/ReceiverModelCard.cs
```diff
@@ -16,15 +17,17 @@ public class ReceiverModelCard : ModelCard
   public bool HasDismissedUpdateWarning { get; set; }
   public List<string>? BakedObjectIds { get; set; }
 
-  public ReceiveInfo GetReceiveInfo(string sourceApplication) =>
-    new(
-      AccountId.NotNull(),
-      new Uri(ServerUrl.NotNull()),
+  public ReceiveInfo GetReceiveInfo(IAccountService accountService, ISpeckleApplication application)
+  {
+    var account = accountService.GetAccountWithServerUrlFallback(AccountId.NotNull(), new Uri(ServerUrl.NotNull()));
```
oguzhankoral, line 22: We should not fallback to server url related account if the from model card. Instead I wanna scream as error on connectors side. Because we already have logics in DUI side to handle these scenarios.

Note for later: DUI related state objects should be as is always in their logics.

DUI3/Speckle.Connectors.DUI/Models/Card/SenderModelCard.cs
```diff
@@ -11,6 +12,9 @@ public class SenderModelCard : ModelCard
   // [JsonIgnore]
   // public HashSet<string> ChangedObjectIds { get; set; } = new();
 
-  public SendInfo GetSendInfo(string hostApplication) =>
-    new(AccountId.NotNull(), new Uri(ServerUrl.NotNull()), ProjectId.NotNull(), ModelId.NotNull(), hostApplication);
+  public SendInfo GetSendInfo(IAccountService accountService, ISpeckleApplication application)
+  {
+    var account = accountService.GetAccountWithServerUrlFallback(AccountId.NotNull(), new Uri(ServerUrl.NotNull()));
```
oguzhankoral, line 17: same comment as above

clairekuang (approved):

PR #953, opened 2025-06-26
fix(grasshopper): block instance transforms incorrect after document unit changes

bjoernsteinhagen:
## Description
Fixes 🐞  where block instance transforms become stale when switching between Rhino documents with different units. When users create block instances in a meters document, then switch to a millimeters document, the transforms retain the old unit assumptions causing incorrect scaling (geometry appears 1000x larger/smaller in the metres <-> millimetres case).

## User Value
Block instances now maintain correct transforms and scaling when switching between documents with different units, preventing geometry corruption during send operations. Don't know how common this is, but at least we can be confident that users won't get funny sends.

## Changes:
- Added `DocChangeMonitor` static class to handle Rhino document change events
- Added `IDocChangeListener` interface for components that need document change notifications  
- Updated `SpeckleBlockInstancePassthrough` to subscribe to document changes and auto-expire

### Before: Block instances sent with incorrect scaling after document unit change

https://github.com/user-attachments/assets/a8d42bb7-8637-46ed-9aa1-ce2712262395

### After: Block instances automatically recalculate and send with correct scaling

https://github.com/user-attachments/assets/1df63076-79f3-43d9-aa92-05dbeef96e5e

## Validation of changes:
Tested reproduction steps:
1. Create block instances in meters document
2. Publish model and inspect
3. Switch to new document in different units
4. Publish again - geometry now maintains correct scale without manual recomputation

**Note:** Component automatically expires and recalculates on document changes

## Checklist:

### DIFF
--- a/Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Components/Objects/SpeckleBlockInstancePassthrough.cs
+++ b/Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Components/Objects/SpeckleBlockInstancePassthrough.cs
@@ -29,7 +29,8 @@ public class SpeckleBlockInstancePassthrough : GH_Component, IDocChangeListener
 
   protected override void RegisterInputParams(GH_InputParamManager pManager)
   {
-    int instanceIndex = pManager.AddGenericParameter(
+    int instanceIndex = pManager.AddParameter(
+      new SpeckleBlockInstanceParam(),
       "Block Instance",
       "BI",
       "Input Block Instance. Speckle instances and Grasshopper instances are accepted.",
@@ -37,7 +38,8 @@ public class SpeckleBlockInstancePassthrough : GH_Component, IDocChangeListener
     );
     Params.Input[instanceIndex].Optional = true;
 
-    int definitionIndex = pManager.AddGenericParameter(
+    int definitionIndex = pManager.AddParameter(
+      new SpeckleBlockDefinitionWrapperParam(),
       "Definition",
       "D",
       "Block Instance Definition. Speckle definitions and Grasshopper definitions are accepted.",
@@ -56,7 +58,8 @@ public class SpeckleBlockInstancePassthrough : GH_Component, IDocChangeListener
     int nameIndex = pManager.AddTextParameter("Name", "N", "Name of the Speckle Instance", GH_ParamAccess.item);
     Params.Input[nameIndex].Optional = true;
 
-    int propIndex = pManager.AddGenericParameter(
+    int propIndex = pManager.AddParameter(
+      new SpecklePropertyGroupParam(),
       "Properties",
       "P",
       "The properties of the Speckle Instance. Speckle Properties and User Content are accepted.",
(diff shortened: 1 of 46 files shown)


Connectors/Rhino/Speckle.Connectors.GrasshopperShared/HostApp/DocChangeMonitor.cs
```diff
@@ -0,0 +1,95 @@
...
+namespace Speckle.Connectors.GrasshopperShared.HostApp;
+
+/// <summary>
+/// Monitors Rhino doc events (creating, opening, switching) and notifies subscribed components when documents change.
+/// These aren't caught by Grasshopper's DocumentContextChanged.
+/// </summary>
+/// <remarks>
+/// Not tied to any specific component - doc events are global. When any doc changes, all interested components should know.
+/// </remarks>
+public static class DocChangeMonitor
```
adamhathcock, line 14: This is an EventAggregator.  Using statics with WeakReferences is fine but I'd rather DI was involved.

I'm more pointing this out because we removed an earlier implementation because of threading issues.

adamhathcock (commented):
Just commenting

bjoernsteinhagen:
@clairekuang, #954 merged in to this. Re-tested and working as shown above. Happy for a review 👍🏼

Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Components/Objects/SpeckleBlockInstancePassthrough.cs
```diff
@@ -17,7 +18,10 @@ public SpeckleBlockInstancePassthrough()
       "Create or modify a Speckle Block Instance",
       ComponentCategories.PRIMARY_RIBBON,
       ComponentCategories.OBJECTS
-    ) { }
+    )
+  {
+    DocChangeMonitor.Subscribe(this);
```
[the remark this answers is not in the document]
bjoernsteinhagen: Probably 5 - 10 components (max) and this happens once per component instance. `WeakReference<T>` to prevent memory leaks. But consdering bottom comment, should we use DI instead?

Connectors/Rhino/Speckle.Connectors.GrasshopperShared/HostApp/DocChangeMonitor.cs
bjoernsteinhagen: Linked to above, i suppose

bjoernsteinhagen:
A couple things here. This is now growing from what I though was a simple improvement:
- Opening a new document with different units won't fire `DocumentPropertiesChanged` (or from what I understand)
- Switching between existing documents with different units won't fire `DocumentPropertiesChanged` (again, from my understanding)
- Only changing units within the current document fires `DocumentPropertiesChanged`

Should we then be targeting all scenarios?

clairekuang:
I see, let's go over this tmr and rubber duck on it. I wouldn't worry too much about this fix because doc/unit switches are probably uncommon for gh users

PR #954, opened 2025-06-27
fix(grasshopper): a bunch of instance vs object casting issues

clairekuang:
## Description
- adds helper methods to create the correct objectwrapper type for speckle objects vs instances
- uses component input and output typing as appropriate for implicit casting
- Removes deep copying in casting. Now all components relying on this should explicitly deep copy wrappers if they are mutating inputs. This should improve any node to node operation that does not mutate objects
- uses generic inputs and outputs for objects and instances. Learning here is that the wrappers just need to be created into their respective goos before setting generic outputs to be registered in their correct type on input

## Checklist:


clairekuang:
@bjoernsteinhagen pushed my last changes, now this is actually ready for review

bjoernsteinhagen (approved):

PR #955, opened 2025-06-27
Feat(gh): sign in component

oguzhankoral:
Triggers desktop service for auth flow.
While auth flow is processing we check number of accounts via account manager in timer to understand something has added or not. It is hacky but thats the only way to understand an account added. If user try to add same account, number of accounts will be same thats why we throw warning after 30 sec as `Sign in timed out. This may have happened because you tried adding an existing account.`

NOTE: we need very talented graphic designer intern for component icon @clairekuang 

rider64_udnD6Rjj0d

Next steps: we can add menu to be able to remove accounts if needed. I believe it is low prio right now


clairekuang:
tested, seems like I get the timeout warnign even after successfully linking an account:
image
this may be because i was already signed in, or initiated the account auth on a non-signed in browser and didn't complete before adding it on another browser.

clairekuang (approved):

PR #956, opened 2025-06-27
Fix(gh): reset menu list on closed according to search text

oguzhankoral:
We reset now list of items by refetching them on closed only if search text is not null.

Rhino_VH7giJoxjc


bjoernsteinhagen (approved):
beautiful ❤️

PR #990, opened 2025-07-18
fix(rhino): replacing invalid chars in block definition names

kekesidavid:
- replacing invalid chars ("/" and "\\") in block definition names with ("-") on receive


clairekuang:
@kekesidavid could you add this as a general helper function to clean both layer and block definition names?

I would include all invalid characters, such as `* : ?` and others.

Connectors/Rhino/Speckle.Connectors.RhinoShared/Operations/Receive/RhinoHostObjectBuilder.cs
```diff
@@ -242,6 +246,16 @@ CancellationToken cancellationToken
     return Task.FromResult(new HostObjectBuilderResult(bakedObjectIds, conversionResults));
   }
 
+  private string ReplaceInvalidChars(string input)
+  {
```
clairekuang, line 250: can you include all invalid characters here, and move this to a general helper function to clean all layer names and block definition names. This currently only fixes the base layer name

clairekuang (approved):

PR #991, opened 2025-07-18
feat(logging): Track a static session id

JR-Morgan:
Many logs in seq lack a user correlating ID, this is because only some functions that log actually know what `Account` is in use.

I have a quick and dirty solution that will help us correlate logs. A static session id that is consistent across all logs from a user's session...

It's not perfect, but is
 - Completely anonymous, no user data involved.
 - Would give us a more accurate way to determine which logs correlate to other related logs in seq


adamhathcock (approved):

JR-Morgan:
closing in favour of https://github.com/specklesystems/speckle-sharp-connectors/pull/997

PR #994, opened 2025-07-22
fix(autocad): converts invalid property value types to string

clairekuang:
## Description

extension dictionary entries can contain values like objectIds that will cause serialisation exceptions.
this pr converts non-simple values to strings.

fix for: https://linear.app/speckle/issue/CNX-2141/sdk-serialisation-exception-from-autocad

## Screenshots:

image
https://app.speckle.systems/projects/b6e95c0c63/models/ff340eed0c

## Checklist:


Converters/Autocad/Speckle.Converters.AutocadShared/ToSpeckle/Properties/ExtensionDictionaryExtractor.cs
```diff
@@ -55,4 +58,21 @@ public ExtensionDictionaryExtractor(IConverterSettingsStore<AutocadConversionSet
 
     return extensionDictionaryDict.Count > 0 ? extensionDictionaryDict : null;
   }
+
+  // xrecord values can contain invalid objects like objectIds
+  // https://help.autodesk.com/view/OARX/2025/ENU/?guid=OARX-ManagedRefGuide-Autodesk_AutoCAD_DatabaseServices_DxfCode
+  private object? GetValidValue(object val) =>
+    val switch
+    {
+      string s => s,
```
JR-Morgan, line 67: Would it be desirable to do the following instead

```cs
private object? GetValidValue(object val)
{
  if(val.GetType().IsPrimitive)
  {
    return val;
  }
  else
  {
    return val.ToString();
   }
}
 ```
 
from [the ms docs](https://learn.microsoft.com/en-us/dotnet/api/system.type.isprimitive):
> The primitive types are [Boolean](https://learn.microsoft.com/en-us/dotnet/api/system.boolean?view=net-9.0), [Byte](https://learn.microsoft.com/en-us/dotnet/api/system.byte?view=net-9.0), [SByte](https://learn.microsoft.com/en-us/dotnet/api/system.sbyte?view=net-9.0), [Int16](https://learn.microsoft.com/en-us/dotnet/api/system.int16?view=net-9.0), [UInt16](https://learn.microsoft.com/en-us/dotnet/api/system.uint16?view=net-9.0), [Int32](https://learn.microsoft.com/en-us/dotnet/api/system.int32?view=net-9.0), [UInt32](https://learn.microsoft.com/en-us/dotnet/api/system.uint32?view=net-9.0), [Int64](https://learn.microsoft.com/en-us/dotnet/api/system.int64?view=net-9.0), [UInt64](https://learn.microsoft.com/en-us/dotnet/api/system.uint64?view=net-9.0), [IntPtr](https://learn.microsoft.com/en-us/dotnet/api/system.intptr?view=net-9.0), [UIntPtr](https://learn.microsoft.com/en-us/dotnet/api/system.uintptr?view=net-9.0), [Char](https://learn.microsoft.com/en-us/dotnet/api/system.char?view=net-9.0), [Double](https://learn.microsoft.com/en-us/dotnet/api/system.double?view=net-9.0), and [Single](https://learn.microsoft.com/en-us/dotnet/api/system.single?view=net-9.0).

clairekuang: ah good point 😅

JR-Morgan (approved):

PR #1003, opened 2025-07-25
feat(revit): extract system type params

kekesidavid:
- `GetMaterialIds()` returns nothing on MEP system types
- the material is stored as a parameter
- currently `ParameterExtractor` only extracts `Instance Parameters` and `Type Parameters`
- the current PR adds `System Type Parameters` as well to include MEP materials

Test Commit:
https://app.speckle.systems/projects/c751bae0f6/models/b12f172e0f

image


clairekuang (changes requested):
This pr would need to extract material quantities (original issue), which means getting the volume calculations per material and adding them to the `materialQuantities` dictionary in properties

clairekuang (approved):
Actually nvm misread the issue - looks like materials are enough!

PR #1007, opened 2025-07-29
Chore(rhino): Quick easy nullability fixes

JR-Morgan:
I was experimenting with multi-targeting rhino.
I'm putting that experiment on hold for the minute, however,

There was several "easy" fixes regarding nullability that we could benefit from having on `dev`.

All the changes I've made should be very safe, if you spot anything that could suspicious flag it up.


adamhathcock (approved):
Proper nullability for the win

PR #1029, opened 2025-08-14
Allows for multiface painting of meshes

jsdbroughton:
Simplifies the process of face addition by introducing a helper method to organize faces by material. 

- Consolidates face addition logic to support consistent material use
- Enhances mesh data conversion implementation to address existing inefficiencies

This includes a refactor stabilizing non-planar quad handling and optimizes n-gon triangulation, ensuring reliable geometric object creation.


libnypacheco:
This is wonderful, Jonathon!

PR #1038, opened 2025-08-21
feat (revit): added category under properties

bimgeek:
## Description
Adds the processed `category` field to the `properties`, while maintaining the existing top-level `category` field.
Related to [2163](https://linear.app/speckle/issue/CNX-2163/add-category-under-properties).

## User Value
We have helper functions in Power BI that deals with properties. Having category field under properties makes interacting with Revit properties much easier as everything is in one place, under properties.

## Changes
- **ClassPropertiesExtractor.cs**: Added `GetCategoryName()` method with DirectShape handling logic and included `category` in properties dictionary
- **RevitElementTopLevelConverterToSpeckle.cs**: Refactored to use shared category processing method, eliminating code duplication

## Result
- Category now appears in both the top-level field and under `properties`
- No breaking changes

# Screenshots
image

## Validation of changes
- Tested with native Revit elements
- Tested with DirectShapes


bimgeek:
From Claire's response:

I'm not a fan of duplicating props / muddying our object model without having a strategy for doing things consistently.

Adding this means now category is found in 4 places: on the object's collection, on the revitobject, inside the properties as a semantic name, and inside the props for revit's internal name. 
For bashboards, it may end up making sense to add categories/types/etc as proxies instead.

Unless this is preventing a key feature in powerbi or users have complained, I would rather us not change the object model until we get more feedback on bashboards

PR #1040, opened 2025-08-22
fix(revit): Exclude host element from view filter when parts are enabled

bimgeek:
## Description
Fixes the issue in Revit view filter where both Parts and their source elements (floors, walls, etc.) were being collected when Parts Visibility was set to "Show Parts Only". This results in overlapping geometry being published to Speckle.

Fixes [CNX-2121](https://linear.app/speckle/issue/CNX-2121/duplicated-geometry-when-parts-visibility-is-on).

## User Value
Users can now publish Parts-divided elements without getting duplicate geometry.

## Changes:
- Added GetSourceElementIdsToExclude() method to identify source elements from Parts
- Modified RefreshObjectIds() to check view's PartsVisibility setting

## Screenshots:
Before

https://github.com/user-attachments/assets/d7c4ad27-1618-4e48-8d9f-e72930cbca35

After

https://github.com/user-attachments/assets/66470e44-22f8-4712-9a71-ef763f05efdd

## Validation of changes:
- Tested with models containing Parts
- Tested with nested parts hiearchies (divide merge parts)
- Confirmed other parts visibility modes work as expected

## Checklist:


Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/Filters/RevitViewsFilter.cs
```diff
@@ -75,8 +77,22 @@ public List<string> RefreshObjectIds()
...
+      using var initialCollector = new FilteredElementCollector(_doc, view.Id);
+      var allElements = initialCollector.ToElements();
+      var idsToExclude = GetSourceElementIdsToExclude(allElements);
+      
+      elementsInView = allElements.Where(e => !idsToExclude.Contains(e.Id));
+    }
+    else
+    {
+      using var viewCollector = new FilteredElementCollector(_doc, view.Id);
+      elementsInView = viewCollector.ToElements();
```
bjoernsteinhagen, line 94: duplicated code in both the `if` and `else` blocks. Can we maybe try extracting common logic to it's own method? Something, like:

```
private IEnumerable<Element> GetFilteredElementsForView(View view)
{
    using var viewCollector = new FilteredElementCollector(_doc, view.Id);
    var allElements = viewCollector.ToElements();
    
    // parts filtering when view is set to show Parts only (and overwrites allElements)
    if (view.PartsVisibility == PartsVisibility.ShowPartsOnly)
    {
        var idsToExclude = GetSourceElementIdsToExclude(allElements);
        return allElements.Where(e => !idsToExclude.Contains(e.Id));
    }
    
    return allElements;
}
```

`RefreshObjectIds` becomes:

```
public List<string> RefreshObjectIds()
{

   // previous logic as is....

    if (view is null)
    {
        return [];
    }

    // use the refactored method, now
    var elementsInView = GetFilteredElementsForView(view);

    // now use common post-processing for both code paths
    var objectIds = elementsInView
        .Where(e => !string.IsNullOrEmpty(e.Name))
        .Select(e => e.UniqueId)
        .ToList();
    
   // remaining code as usual ...
}

```

Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Send/Filters/RevitViewsFilter.cs
```diff
@@ -125,4 +141,37 @@ public void SetContext(RevitContext revitContext)
...
     _doc = _revitContext.UIApplication?.ActiveUIDocument.Document;
   }
+  
+  // NOTE: Element collector returns parts and source elements even when Parts Visibility is set as "Show Parts" only.
+  // Below function collects list of ids to exclude from final list.
+  private HashSet<ElementId> GetSourceElementIdsToExclude(IEnumerable<Element> elements)
+  {
+    var elementsToExclude = new HashSet<ElementId>();
+
+    foreach (var element in elements)
```
bjoernsteinhagen, line 151: I was initially weary about additional `foreach` loop. Tested LINQ expressions and other native Revit API calls, but this method remained most performant! Let's keep as it, nice!

bjoernsteinhagen (changes requested):
- Refactor to avoid code duplication
- Fix build errors?

bjoernsteinhagen (approved):
looks good :)

PR #1043, opened 2025-08-26
refactor(rhino-importer): Run rhino as a sub process to avoid file locking and improve handling of crashes

JR-Morgan:
Refactor of the job processor and rhino importer to run the rhino importer as a sub-process (rather than running everything in a single process.

There are several bugs this will directly address:
 - https://linear.app/speckle/issue/CNX-2245/importing-3dm-files-causes-fatal-application-crash-on-disposing-of-the
 - https://linear.app/speckle/issue/CNX-2358/locked-files-bug

Additionally, this also should allow for fewer opportunities for data to bleed between imports, since we'll use unique rhino sessions for each import.

---

Also sneaked into this PR some tweaks to the import reader configs for Obj and 3dm to fix the following issues
 - https://linear.app/speckle/issue/CNX-2431/obj-import-settings-need-tweaks-for-y-z-up
- https://linear.app/speckle/issue/CNX-2432/preserve-the-original-units-of-the-3dm-file


JR-Morgan:
Change is currently deployed to https://latest.speckle.systems

Importers/Rhino/Speckle.Importers.JobProcessor/Blobs/ImportJobFile.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+      logger.LogError(ex, "Failed to cleanup file");
+      throw;
+    }
+  }
+
+  ~ImportJobFile()
+  {
+    // Using the full unmanaged style disposal pattern means that the file will be deleted on finalise if dispose wasn't called already.
+    DeleteFile();
```
JR-Morgan, line 47: Probably over kill, and not needed so much now Rhino is a seperate process... but, to provide some context, this is a standard C# pattern around "unmanaged" (i.e. c/c++) style disposal.
Ensuring that, even if `.Dispose` is not called (like in the case of a low level app crash), then the Finalizer will clean up any low level unmanaged resources.

Here, I'm not actually cleaning up any low level unmanaged resources, I'm deleting a file. But I figure that deleting this file is very important, so my hope is that this will be a belts + braces way to ensure it is.

Importers/Rhino/Speckle.Importers.Rhino/Internal/IPCModels.cs
```diff
@@ -0,0 +1,20 @@
...
+  public string ProjectId { get; init; }
+  public string ModelId { get; init; }
+  public Account Account { get; init; }
+}
+
+public readonly struct ImporterResponse
+{
+  public Version? Version { get; init; }
+  public string? ErrorMessage { get; init; }
+}
```
JR-Morgan, line 20: These two structs are copy pasted between the two projects.
A little nasty, but the alternatives are also not too pretty
- 🤪 These live somewhere else (like the SDK or Speckle.Connector.Common, or a new project... pretty nasty) 
- 🤮 We use dynamic definitions (infinitely worse than this)
- 🤷 We have a direct reference JobProcessor -> RhinoImporter... perhaps not too nasty, but I wanted to avoid coupling them directly.

Happy to consider the latter if these two structs offend ppl living in two places.

Importers/Rhino/Speckle.Importers.Rhino/Internal/FileTypeConfig/Rhino3dmConfig.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Speckle.Importers.Rhino.Internal.FileTypeConfig;
+
+/// <summary>
+/// <see cref="RhinoDoc.OpenHeadless(string)"/> will preserve the units defined by the file
+/// </summary>
+public sealed class Rhino3dmConfig : IFileTypeConfig
+{
+  public RhinoDoc OpenInHeadlessDocument(string filePath)
+  {
+    RhinoDoc? doc = RhinoDoc.OpenHeadless(filePath);
```
JR-Morgan, line 13: Using `OpenHeadless` rather than `CreateHeadless` + `Import` preserves units for 3dm files

Importers/Rhino/Speckle.Importers.Rhino/Internal/FileTypeConfig/ObjConfig.cs
```diff
@@ -0,0 +1,39 @@
...
+public sealed class ObjConfig : IFileTypeConfig
+{
+  private readonly FileObjReadOptions _readOptions =
+    new(new FileReadOptions() { OpenMode = true })
+    {
+      MapYtoZ = true,
+      UseObjGroupsAs = FileObjReadOptions.UseObjGsAs.ObjGroupsAsLayers,
+      UseObjObjectsAs = FileObjReadOptions.UseObjOsAs.ObjObjectsAsObjects,
+      IgnoreTextures = true, //We don't support MTL uploads, so no point trying to find a MTL file...
+    };
```
JR-Morgan, line 16: Obj import settings means
1. Obj groups become rhino layers
2. We correctly map Y -> Z up
Before:
image

After:
image

PR #1054, opened 2025-08-28
fix(revit): preserve selection intent while maintaining dynamic filter behavior

bjoernsteinhagen:
## Description

"Fixes" selection filter by persisting original user selection intent for selection filters while maintaining current behavior for categories/views filters. In an attempt to fix a state mismatch where the UI showed live selection count but operations used filtered model reality.

The bug manifested as:
- Model card displaying "3 objects selected" but highlight only showing 2 objects
- Selection filters losing track of originally selected elements after delete+republish+undo workflows  
- Stale parameter data in viewer despite Update button indicating changes

Root cause: `SetFilterObjectIds` was sending filtered object lists (only existing elements) to the frontend for ALL filter types. This overwrote selection filters' `selectedObjectIds` with reduced sets, breaking the original selection intent. However, categories/views filters expect this behavior since they represent dynamic queries ("all walls" vs "these specific columns").

The fix differentiates filter behavior:
- **Selection filters**: Preserve original `SelectedObjectIds` to maintain user intent
- **Categories/Views filters**: Use `newSelectedObjectIds` since they self-update via `RefreshObjectIds()`

Fixes [CNX-2400](https://linear.app/speckle/issue/CNX-2400/object-dont-update-on-publish)

## User Value

Selection filters now properly preserve the original user selection intent while categories/views filters maintain their expected dynamic behavior.

## Changes:

- Added conditional logic in `RevitSendBinding.RefreshElementsIdsOnSender()` to handle selection filters differently
- Selection filters now send their original `SelectedObjectIds` instead of filtered `newSelectedObjectIds`
- Categories/Views filters maintain existing behavior with `newSelectedObjectIds`

### Before

https://github.com/user-attachments/assets/d49ee6b1-c5a6-4f28-adf9-c22170c6a090

### After

https://github.com/user-attachments/assets/35e04e2d-0c25-4cf4-ab37-934e56964c13

## Checklist:


oguzhankoral (commented):
I am soooo skeptical about this removal, we need to make sure all other behaviors will be as expected.

oguzhankoral (approved):

PR #1058, opened 2025-08-29
fix(rhino): Workaround for rhino materials being null in importer

JR-Morgan:
In cases where there's no `RenderMaterial` on a rhino object, but there is a `Material` applied to a geometry we use the `RenderMaterial.FromMaterial` function to create a new `RenderMaterial` from that `Material` to perform conversion.
But for some reason, this function returns null consistently for our importer.
Perhaps it's because converting things from a headless rhino. Or from windowsless rhino (where some rendering engine parts might not be initalised)

This feels like a bug in Rhino. So I'm tempted to report it to McNeel.

However, I have found this workaround that might work nicely for us.

Using `RenderMaterial.CreateBasicMaterial` as fallback appears to work for the importer.

I'll be honest, I'm not sure I understand the difference between this function and `RenderMaterial.FromMaterial`, so there's a chance this does give us a slightly different result. But visually the materials I tested end up looking the same.


JR-Morgan:
Now deployed ready for testing

PR #1059, opened 2025-09-01
fix(sdk) move the logging addition for connectors

adamhathcock:


adamhathcock:
Ohh I forgot that's how we set levels....will fix further

JR-Morgan (approved):

PR #1060, opened 2025-09-01
chore(fileImport): Jedd/cnx 2444 rename rhino file importer to something else

JR-Morgan:
image


bimgeek (approved):
nice.

adamhathcock:
This should be named something less generic.  We have multiple file importers (ACC, ODA, etc.)

I guess we need to drawing a difference between what's logged/managed and what a user sees?

PR #1064, opened 2025-09-03
fix(Revit) - Getting a key on exception could result in null reference exceptions

adamhathcock:
Maybe a fix for:
image


adamhathcock:
Also this:
image

JR-Morgan (approved):

PR #1065, opened 2025-09-03
fix(Revit) - Never use First() as it can be missing

adamhathcock:
Better error than this...can still be null?
image


adamhathcock:
Also this
image

PR #1071, opened 2025-09-04
Avoid using blam operator to ignore nulls

adamhathcock:
Refactored a bit avoid nulls and maybe this exception
image

### DIFF
--- a/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
+++ b/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
@@ -236,17 +236,16 @@ internal sealed class RevitSendBinding : RevitBaseBinding, ISendBinding
   private void DocChangeHandler(Autodesk.Revit.DB.Events.DocumentChangedEventArgs e)
   {
     ICollection<ElementId> modifiedElementIds = e.GetModifiedElementIds();
-
+    var doc = e.GetDocument();
+    if (doc == null)
+    {
+      return;
+    }
     // NOTE: Whenever we save data into file this event also trigger changes on its DataStorage.
     // On every add/remove/update model attempt triggers this handler and was causing unnecessary calls on `RunExpirationChecks`
     // Re-check it once we implement Linked Documents
     if (modifiedElementIds.Count == 1)
     {
-      var doc = e.GetDocument();
-      if (doc == null)
-      {
-        return;
-      }
       if (modifiedElementIds.All(el => doc.GetElement(el) is DataStorage))
       {
         return;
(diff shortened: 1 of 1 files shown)


Connectors/Revit/Speckle.Connectors.RevitShared/HostApp/ElementUnpacker.cs
```diff
@@ -52,8 +41,12 @@ public IEnumerable<Element> UnpackSelectionForConversion(IEnumerable<Element> se
   /// </remarks>
   public IEnumerable<string> GetUnpackedElementIds(IEnumerable<string> objectIds)
   {
-    var doc = _revitContext.UIApplication?.ActiveUIDocument.Document!;
-    var docElements = doc.GetElements(objectIds);
+    var doc = revitContext.UIApplication?.ActiveUIDocument.Document;
+    var docElements = doc?.GetElements(objectIds);
+    if (docElements == null)
+    {
+      return [];
```
JR-Morgan, line 48: Should we not throw if `Document` is null... the function didn't really Unpack the given Element ids... 
sounds like an exceptional case to me...

Connectors/Revit/Speckle.Connectors.RevitShared/HostApp/ElementUnpacker.cs
```diff
@@ -127,11 +123,17 @@ private List<Element> UnpackElements(IEnumerable<Element> elements, Document? do
   // 2- Evicting the cache while introducing the settings
   private List<Element> PackCurtainWallElementsAndStackedWalls(List<Element> elements, Document? doc = null)
   {
-    var ids = elements.Select(el => el.Id).ToArray();
+    var ids = elements.Select(el => el.Id).ToHashSet();
```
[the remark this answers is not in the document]
adamhathcock: It was doing just Contains checks below so I optimized for that

Connectors/Revit/Speckle.Connectors.RevitShared/HostApp/ElementUnpacker.cs
bjoernsteinhagen: Happy to chat on this. I've asked myself the same thing. Validating that `doc` is not `null` seems to happen a lot, like beforehand in the `RevitSendBinding`:

```
    var activeUIDoc =
      _revitContext.UIApplication.NotNull().ActiveUIDocument
      ?? throw new SpeckleException("Unable to retrieve active UI document");
```

or in the `RevitConversionSettingsFactory`:

`    var document = revitContext.UIApplication.NotNull().ActiveUIDocument.Document;`

That's possibly why it seems to be "ignored" here.

bjoernsteinhagen (approved):
Testing looks good 😄

PR #1072, opened 2025-09-04
fix(grasshopper): QuerySpeckleObjects crashes with type-specific outputs when path is set

bjoernsteinhagen:
## Description

Fixed an index out of range exception that occurred when users clicked the "+" button to add type-specific outputs (Block Instances, Points, etc.) to the Query Speckle Objects component **while having a path input connected**.

**Note:** this was never working when users piped in a path input.

## User Value
Users can now use type-specific outputs on Query Speckle Objects with collection paths without errors.

## Changes:
- Replaced `Filters[i - 1]` array indexing with safe param name parsing
- Fixed type-specific outputs to correctly use path-filtered objects instead of the entire collection
- Simplified output logic while maintaining the original code structure

### Before
Component goes red with "Index was out of range" error when adding type-specific outputs **with path input**

image

### After
Component works correctly, showing filtered objects for each type-specific output. Type-specific properties count also shows that we're respecting the filter on the objects.

after-fix

## Checklist:

### DIFF
--- a/Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Components/Objects/QuerySpeckleObjects.cs
+++ b/Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Components/Objects/QuerySpeckleObjects.cs
@@ -136,13 +136,22 @@ public class QuerySpeckleObjects : GH_Component, IGH_VariableParameterComponent
     for (int i = 0; i < Params.Output.Count; i++)
     {
       // determine output values based on parameter type
-      List<SpeckleWrapper> outputValues =
-        i == 0
-          ? filteredObjects
-          : Enum.TryParse(Params.Output[i].Name, out ObjectType filterType)
-          && _filterDict.TryGetValue(filterType, out var filteredList)
-            ? filteredList.Cast<SpeckleWrapper>().ToList()
-            : new List<SpeckleWrapper>();
+      List<SpeckleWrapper> outputValues;
+      if (i == 0)
+      {
+        outputValues = filteredObjects;
+      }
+      else if (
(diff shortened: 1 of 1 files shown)


bjoernsteinhagen:
> image I don't think topology should apply only on the first output

hmm yeah 😞. This was a reason for the error. The topology string expects a specific item count that doesn't match filtered results, causing `ArgumentOutOfRangeException`. The existing `CreateDataTreeFromTopologyAndItems` helper isn't designed for sparse data - it expects the exact object count/order that the topology encodes.

I think the fix may be a little tricky - but I'll look to deeper. Worth the complexity, or we stick with flat lists for filtered outputs?

Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Components/Objects/QuerySpeckleObjects.cs
```diff
@@ -136,10 +135,19 @@ protected override void SolveInstance(IGH_DataAccess dataAccess)
     // Set output objects
     for (int i = 0; i < Params.Output.Count; i++)
     {
+      // determine output values based on parameter type
       List<SpeckleWrapper> outputValues =
-        i == 0 ? filteredObjects : _filterDict[Filters[i - 1]].Select(o => (SpeckleWrapper)o).ToList();
-      List<IGH_Goo> outputGoos = outputValues.Select(o => o.CreateGoo()).ToList();
-      if (targetCollectionWrapper?.Topology is string topology && !string.IsNullOrEmpty(topology))
+        i == 0
```
adamhathcock, line 140: can you read this ternary operator mess?

I recommend redoing it to just have if statements.

clairekuang (approved):

PR #1082, opened 2025-09-10
Refactor(metrics): Refactor metrics tracking to avoid bad defaults

JR-Morgan:
There was a lot of legacy fallback behaviour that makes the metrics tracking fragile


JR-Morgan:
needs a quick test in grasshopper,
It looks like we're always expecting a non-null account in all the usages, so we should safe. But want to do a quick test anyway.

adamhathcock (dismissed):
less code is more code

JR-Morgan:
Tested in grasshopper, no issues with the normal nodes.

I've made one small breaking change to the dev `TokenUrlComponent` to require the email scope.
Without this, there's no accurate way to track metrics for sends/receives that use this token....

I've checked with @clairekuang, and this should be fine. The component as is already requires `profile:read`, `stream:read`, and `workspace:read` scopes...

PR #1085, opened 2025-09-10
fix(revit): invert exception filter condition in ClassPropertiesExtractor

bjoernsteinhagen:
## Description
Fixed inverted exception filter in `ClassPropertiesExtractor` - was catching fatal exceptions instead of non-fatal ones, causing sends to crash on linked model elements.

**Potential NullRef sources:**
- `element.Document.GetWorksetTable()` returns null (non-workshared linked doc)
- `element.WorksetId` is null when passed to `GetWorkset()`

## User Value
Send operations no longer crash when processing elements from linked models.

## Changes:
- Fixed exception filter condition in `ClassPropertiesExtractor.ExtractElementProperties()`
- Now catches `NullReferenceException` and provides "Unknown Workset" fallback

## Checklist:


clairekuang (approved):
oof

PR #1094, opened 2025-09-15
fix(rhino): units scaling for text entities

JR-Morgan:
While investigating other text related issues, I discovered issues receiving text into a document with a different unit from the original document.
image

---

Our normal unit scaling logic in rhino `ToHost` functions relies on a `GeometryBase.Transform` function call in the  `SpeckleToHostGeometryBaseTopLevelConverter` super class.

However, it looks like said `Transform` function (internal name `ON_Geometry_Transform`) does not behave the way we expected for Text entities.
It doesn't appear to scale the text height (despite it being linear units)

Interestingly `TextEntity` does have a different `Transform` overload that does appear to change text height
[`Transform(Transform, DimensionStyle)`](https://github.com/mcneel/rhino3dm/blob/71c63a8c1c87782a13a1b76c825e4b792b36fd09/src/dotnet/opennurbs/opennurbs_text.cs#L270-L289) (internal name `ON_V6_TextObject_Transform`)

---

In this PR, I've just added simple units scalling directly in the TextEntity raw converter. This breaks the pattern slightly, but given the API, I think it's the cleanest option.


bjoernsteinhagen (approved):
🫡

PR #1101, opened 2025-09-18
fix (Revit) Check if the connector is active before using

adamhathcock:
Trying to stop this error:
image


bjoernsteinhagen (commented):
Stale? #1109

adamhathcock:
Yes, I knew I had done this before.

PR #1118, opened 2025-09-29
chore: Remove legacy .NET based IFC importer

JR-Morgan:
.NET importer is no longer being deployed or used.

This PR removes it from the code base.


JR-Morgan:
need to decide what to do r.e. nuget CI

PR #1121, opened 2025-09-30
fix (revit) Use current doc as soon as possible instead of delegating the grab further down

adamhathcock:
Attempting to fix:
image


Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
```diff
@@ -110,6 +109,11 @@ public List<ICardSetting> GetSendSettings() =>
 
   public async Task Send(string modelCardId)
   {
+    var document = _revitContext.UIApplication?.ActiveUIDocument?.Document;
+    if (document == null)
+    {
+      return;
```
JR-Morgan, line 115: We should throw here. I don't like the idea that a function called "Send" can silently returns without actually sending...  

The send failed, the document state was invalid... exception.

JR-Morgan (changes requested):

JR-Morgan (approved):
tested revit send

PR #1122, opened 2025-09-30
feat(nuget): Publish Speckle.Connectors.Logging as nuget

JR-Morgan:
wip refactor

PR #1123, opened 2025-09-30
chore(ci): Refactor out nuget publishing from CI, use linux for pr builds

JR-Morgan:
A few months ago, #998 was my first attempt to switch the CI builds to using linux hit a road block due to the way we publish nugets.

Then today, I came to do #1122  to publish Speckle.Connectors.Logging as nuget. So I started a refactor, but since having a chat with Adam, this isn't a requirement for us to do logging in the desktop service. 
The nuget we publish from this repo was the old .NET based IFC importer.

Thus this PR is just the following refactor of the CI:
1. Switch the PR workflow to run on Linux runners. This reduces the CI time for PRs from ~5m -> ~3m30s. 
    I've kept the release pipeline on windows since cross platform building of exes requires us to use `dotnet publish`...
2. Renamed some of the build commands to make it clearer which only operate on affected projects vs all, and to avoid now ambiguious "-linux" names...
3. Disabled the nuget publishing step, since we no longer require any nugets to be pushed from this repo. However, I've left the code there ready in case we need to in future.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -14,6 +14,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
 
       - name: Setup .NET
         uses: actions/setup-dotnet@v5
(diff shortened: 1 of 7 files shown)


JR-Morgan:
Low prio on this @adamhathcock 
not blocking anything

adamhathcock (approved):

PR #1126, opened 2025-10-01
feat(autocad/civil): add reference point behavior

clairekuang:
## Description

Sends all objects according to the current selected User Coordinate System.
Previously, models were published according to the WCS (world coordinate system) regardless of the current UCS, and some conversions were incorrectly transformed due to certain geometries storing values in UCS instead of WCS.
This pr fixes publishes from autocad and civil when the current CS is set to UCS, by staying 1-1 with how the objects are oriented in-app.

Major changes to review:
- Speckle.Connectors.Common.Instances changes
- breaking change in polyline to speckle conversion: value list is changed from point2d to point3d! need to reconsider this

Also includes the following minor unrelated changes:
- no longer sending the default render material if an object or layer does not have a render material assigned.
- all bbox conversions removed

## User Value

Enables smooth reference point coordination workflows between autocad/civil3d and other connectors. Was frequently reported by civil users struggling to coordinate with revit.

Process before:
- precoordinate survey point in civil and in revit
- publish from civil (using wcs)
- move revit survey point to origin
- receive in revit according to survey point

Process after:
- precoordinate survey point in civil and in revit
- publish from civil (using correct UCS)
- receive in revit according to survey point

## Changes:

- autocad geometry conversions (incl fixes to currently failing classes)
- added reference point helpers in autocad converter
- civil object conversions
- modified connectors common instance interfaces to identify block definition objects separately from atomic objects
- note: does NOT add current UCS to root collection on publish - this is because revit will undo transforms if it detects one.
- modified send operation so that: (1) definition objects are NOT transformed in the converter (2) top-level instance transforms are modified by the current UCS

## To-do before merge:

- [x] Check with @oguzhankoral on connectors common instance implementation to distinguish definition objects
- [ ] test fully civil -> revit workflows
- [x] add autocaddocstore to track current UCS/WCS changes to invalidate cache (#1164)
- [x] fix rotated polyline receive issues in autocad
- [x] resolve breaking change in polyline conversion (see above)
- [x] fix tangent receive issues in polylines

## Screenshots:

**Test file: Autocad Objects test file**
image

**Before:**
Sent according to UCS: https://app.speckle.systems/projects/b6e95c0c63/models/2fda739bc3@b51195dfd4
image

Received version in Rhino:
image

**After:**
Sent according to WCS: https://app.speckle.systems/projects/b6e95c0c63/models/beccdd02f8@1371c20584
Sent according to UCS: https://app.speckle.systems/projects/b6e95c0c63/models/e3cba24d1f@2f7c15bb93

image

Original file in UCS vs received UCS version in Autocad:
499963143-dcd40768-6633-4531-910f-38f2cd481b11

**note**: polyline2d tangent values are not being converted back to native correctly. need to fix this.

Received UCS version in Rhino:
image

Received UCS version in Revit:
image

## Checklist:

### DIFF
--- a/Converters/Revit/Speckle.Converters.RevitShared/Helpers/DisplayValueExtractor.cs
+++ b/Converters/Revit/Speckle.Converters.RevitShared/Helpers/DisplayValueExtractor.cs
@@ -68,6 +68,14 @@ public sealed class DisplayValueExtractor
         return [DisplayValueResult.WithoutTransform(GetCurveDisplayValue(modelCurve.GeometryCurve))];
       case DB.Grid grid:
         return [DisplayValueResult.WithoutTransform(GetCurveDisplayValue(grid.Curve))];
+      case DB.Architecture.Room room:
+        // api still returns geometry for unplaced rooms.
+        // return empty list so room object is sent but with null display value
+        if (room.Volume == 0)
+        {
+          return new List<DisplayValueResult>();
+        }
+        return GetGeometryDisplayValue(room);
       case DB.Area area:
         return _converterSettings.Current.SendAreasAsMesh
           ? GetAreaMeshDisplayValue(area)


Connectors/Autocad/Speckle.Connectors.AutocadShared/Operations/Send/AutocadRootObjectBaseBuilder.cs
```diff
@@ -104,21 +125,53 @@ CancellationToken cancellationToken
...
+          projectId,
+          referenceTransform // set this for top level instance proxies to use if needed
+        );
+
         results.Add(result);
 
         onOperationProgressed.Report(new("Converting", (double)++count / atomicObjects.Count));
       }
 
+      // 4 - Convert atomic definition objects
```
clairekuang, line 142: @oguzhankoral this is the impact of two separate atomic object lists from the instance unpacker, looks bad imo

bimgeek (approved):
I've tested with multiple models. It's looking good. There is an issue when loading with UCS in Civil 3D. we logged that as a separate item and will merge this one in. It's been in the backlog for a while.

PR #1128, opened 2025-10-02
cache parameters by group and name instead of caching groups and assuming units are all the same

adamhathcock:


Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/Properties/ParameterDefinitionHandler.cs
```diff
@@ -5,13 +5,15 @@ namespace Speckle.Converters.RevitShared.ToSpeckle;
...
+  private sealed record ParameterDefinition(string GroupName, string? Units);
+
+  private sealed record ParameterKey(string InternalName, string Group);
 
   /// <summary>
   /// Keeps track of all parameter definitions used in the current send operation. This should be attached to the root commit object post conversion.
   /// </summary>
   /// POC: Note that we're abusing dictionaries in here because we've yet to have a simple way to serialize non-base derived classes (or structs?)
-  private readonly Dictionary<string, GroupDefinition> _groupDefinitions = new();
+  private readonly Dictionary<ParameterKey, ParameterDefinition> _parameterDefinitions = new();
```
JR-Morgan, line 16: Because `record` types override `.Equals` and `.GetHasCode`, it is safe to use them as key right?

adamhathcock: yup!

bjoernsteinhagen (approved):
confirm problematic model now reportig correctly

PR #1130, opened 2025-10-02
Key for docs should fallback correctly and load when not found.

adamhathcock:
Revit 24+ properly uses creation guid which is unique per doc instance.
Revit 22/23 uses path (if it exists) for a best effort for cards.

A null key or null state will clear out the current cards.


bjoernsteinhagen (approved):
💯 .. and the comments ... beautiful

PR #1137, opened 2025-10-06
fix(rhino-importer): Return non-zero exit code on fault

JR-Morgan:
Exceptions that occur during the execution of a background service to not surface to the apphost `RunAsync` which would allow them to terminate the process with a non-zero exit code.
See related discussion https://github.com/dotnet/runtime/issues/67146


JR-Morgan:
tested locally

PR #1138, opened 2025-10-06
Stop exceptions we can't do anything about

adamhathcock:
This probably is a random Revit issue we can't do anything about

stops

image


JR-Morgan (commented):
Revit/cef seem to not be the kindest with these edge cases.
I'm up for any change that means we get less spammed logs.

PR #1140, opened 2025-10-07
Feat: display value proxies

oguzhankoral:


Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/Raw/Geometry/MeshByMaterialDictionaryToSpeckle.cs
```diff
@@ -85,32 +91,37 @@ public MeshByMaterialDictionaryToSpeckle(
...
         : _converterSettings.Current.Document.GetElement(materialId) is DB.Material material
           ? _speckleRenderMaterialConverter.Convert(material)
           : null;
 
-      // get the render material if any
+      // Create proxy but DON'T populate objects list yet
       if (renderMaterial is not null)
       {
-        if (!materialProxyMap.TryGetValue(materialIdString, out RenderMaterialProxy? renderMaterialProxy))
+        if (!materialProxyMap.TryGetValue(materialIdString, out _))
```
JR-Morgan: Could we use .Contains here?

PR #1150, opened 2025-10-15
fix(grasshopper): account selection not respected when url input connected

bjoernsteinhagen:
## Description
Fixed a bug where changing accounts via right-click context menu didn't work when a URL input was connected to the `SpeckleSelectModelComponent`. The component would revert back to a "default" account after `SetAccount()` triggered `ExpireSolution(true)`.

Root cause: `SolveInstanceWithUrlInput()` was **always** re-deriving and overwriting the account from the URL, ignoring user selection.

Fixes [CNX-2688](https://linear.app/speckle/issue/CNX-2688/changing-accounts-linked-to-the-same-server-on-load-component).

## User Value
Users can now switch between multiple accounts for the same server when using URL inputs (e.g., two accounts both accessing `speckle.systems`), allowing them to control which credentials are used for a given model.

## Changes:
- **Refactored account restoration logic** into `RestoreAccountFromStoredState()` method to eliminate code duplication between SCENARIO 1 (URL input connected) and SCENARIO 2 (manual selection)
- **Modified `SolveInstanceWithUrlInput()`** to only set account when `SelectedAccount == null`, respecting user's explicit account selection via right-click
- **Added server validation** to throw clear error when selected account doesn't match URL's server
- **Added restoration call in SCENARIO 1** before URL resolution to ensure pasted/loaded components restore saved account before URL processing

### Before

https://github.com/user-attachments/assets/8ca6b7b3-4c6b-43f3-89b0-6c906bcf6327

### After

https://github.com/user-attachments/assets/a525d9f7-82b2-424b-9ab2-21999f2d6f18

The video shows:
- Fresh component with URL → auto-selects account from URL ✅
- Right-click account change → uses new account, doesn't revert ✅
- Server mismatch → throws descriptive error ✅
- Multiple accounts for same server → allows switching ✅
- Paste/load component → restores saved account from serialized state ✅

## Checklist:

### DIFF
--- a/Build/packages.lock.json
+++ b/Build/packages.lock.json
@@ -16,12 +16,12 @@
       },
       "Microsoft.Build": {
         "type": "Direct",
-        "requested": "[17.11.4, )",
-        "resolved": "17.11.4",
-        "contentHash": "UMC7DfeFEHY2GGHHaghybUuUlLaByFHEFudR2PehMgDBuRuLAUePp1iaa4eFtVzepRzMtIbeSCVJCzzX3NV2Gg==",
+        "requested": "[17.11.48, )",
+        "resolved": "17.11.48",
+        "contentHash": "g8Kn575mNAKcuFotV3C7xvF+IbxuHennl67LH2shL2au1U6UqwReTDygCHyU04+koc2Yn7fHIbVQaC08HqEWow==",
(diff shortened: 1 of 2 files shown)


JR-Morgan:
🏎️

PR #1159, opened 2025-10-21
fix(grasshopper): applicationId for SpeckleDataObject on creation

bjoernsteinhagen:
## Description

Fixes [CNX-2699](https://linear.app/speckle/issue/CNX-2699/index-was-out-of-range-exception-on-publish-load-workflows).

## User Value

Expected behaviour for publish-load workflows (no "disappearing" objects on load). Create collection provides safeguard and warnings when same object added multiple times to same collection.

## Changes:

- `SpeckleDataObjectPassthrough` generates `ApplicationId` when creating new instance
- `CreateCollection` validates provided objects and checks that there aren't duplicates

## Validation of changes:

- Attahced user script (from linear ticket) now works as expected
- Try adding same object multiple times to create collection throws error on component

## Checklist:


Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Components/Collections/CreateCollection.cs
```diff
@@ -64,6 +64,13 @@ protected override void SolveInstance(IGH_DataAccess dataAccess)
       }
     }
 
+    // validate for duplicate application IDs across the entire collection hierarchy
+    if (HasDuplicateApplicationIds(rootCollection))
```
bjoernsteinhagen: negligible impact on solve times (tested with loading big revit model and creating collection from received model)

clairekuang (changes requested):
Just add a test case for dataobjects created with the same speckle geometry

bjoernsteinhagen:
> Just add a test case for dataobjects created with the same speckle geometry

Taking the problematic model, casting the point to speckle geo and then checking count published versus count received :)

image

clairekuang (approved):

PR #1161, opened 2025-10-21
fix(rhino-importer): Use main thread always for document creation

JR-Morgan:
Occasionally, we're seeing `OpenHeadless` crash the importer with the following error
```
Fatal error.
System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
Repeat 2 times:
--------------------------------
   at UnsafeNativeMethods.CRhinoDoc_OpenHeadless(System.String, IntPtr)
--------------------------------
   at Rhino.RhinoDoc.OpenHeadless(System.String)
Process finished with exit code -1,073,741,819.
```

This was leading to many 3dm imports failing with the error message "`Importer left an invalid response`"

`OpenHeadless` appears to behave un-reliably when used from not the main thread. No idea why we're only noticing this now, could be some async stuff we've tweaked, could be an update to Rhino, not sure.

I'm pretty sure this is a bug, and Dale from mcneel has tracked it - https://discourse.mcneel.com/t/rhino-inside-fatal-app-crashes-when-disposing-headless-documents/208673/6

---

For the mean time, I believe we can work around the issue by ensuring we're always calling `OpenHeadless` from the main thread. This isn't super easy to pull off, because Rhino will also hog the main thread if it's ever yielded, so this can only be done before any `await` keyword is used.
And there's no compile time safety here, we just have to document and remember not to do any async stuff until after opening the document.

### DIFF
--- a/Importers/Rhino/Speckle.Importers.Rhino/Program.cs
+++ b/Importers/Rhino/Speckle.Importers.Rhino/Program.cs
@@ -58,7 +58,7 @@ public static class Program
         {
           try
           {
-            Version result = await importer.RunRhinoImport(importerArgs, CancellationToken.None).ConfigureAwait(false);
+            Version result = await importer.RunRhinoImport(CancellationToken.None).ConfigureAwait(false);
             WriteResult(new() { Version = result }, importerArgs.ResultsPath);
           }
           catch (Exception ex)
--- a/Importers/Rhino/Speckle.Importers.Rhino/Internal/ImporterInstance.cs
+++ b/Importers/Rhino/Speckle.Importers.Rhino/Internal/ImporterInstance.cs
@@ -28,7 +28,7 @@ internal sealed class ImporterInstance(ImporterArgs args, Sender sender, ILogger
     UserActivityScope.AddUserScope(args.Account),
   ];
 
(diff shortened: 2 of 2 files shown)


Importers/Rhino/Speckle.Importers.Rhino/Program.cs
```diff
@@ -32,13 +37,34 @@ public static async Task Main(string[] args)
...
 
-      var importer = serviceProvider.GetRequiredService<ImporterInstance>();
+      var factory = serviceProvider.GetRequiredService<ImporterInstanceFactory>();
+
+      // Error handling flow below here looks a bit of a mess, but we're having to navigate threading issues with rhino inside
+      // https://discourse.mcneel.com/t/rhino-inside-fatal-app-crashes-when-disposing-headless-documents/208673/7
+      try
+      {
+        // This needs to be called on the main thread
+        importer = factory.Create(importerArgs);
```
JR-Morgan: This is the main change. We're opening the document here on construction of the importer.
This way, we're doing it on the main thread.

Previously, this was done from a .NET worker thread inside the `Task.Run` block lower down.

PR #1332, opened 2026-03-23
feat(grasshopper): accept a mix of data objects and collections

bjoernsteinhagen:
## Description

The data structure we publish often has a mixture of `SpeckleDataObject` (and/or `SpeckleGeometry`) and `Collection`. GH invalidates it. So this strict validation was just an artificial blocker. 

This PR removes that constraint so users can pipe in a mixture of data objects, speckle geometries, and collections. We also updated the Send components so if you pipe a mixed list directly into them, it auto-wraps everything into a root collection.

Completes [CNX-3175](https://linear.app/speckle/issue/CNX-3175/collections-node-should-accept-a-mix-of-data-objects-collections).

## User Value

Users can now combine data objects and collections using standard GH merge components without hitting errors, and can even skip the `Create Collection` node entirely for simple scripts.

## Changes:

- Reworked `CreateCollection.cs` to process mixed inputs in a single loop
- Updated `SendComponent.cs` and `SendAsyncComponent.cs` inputs to accept generic lists instead of a single collection item.
- Added an root builder in the Send components to auto-wrap mixed inputs 

- **Green**: shows piping data objects straight in to publish component. Wrapped in root and published
- **Red**: shows collection containing mixture of `DataObject` and child `Collection`.
- **Blue**: shows expanding preserving inputs.
 
Screenshot 2026-03-23 160539

No changes were needed on the receive/expand workflows as the core `Base` and Grasshopper unpackers already supported mixed elements.

## Before Merge:
- [ ] @bimgeek I renamed `Collections` on the publish component to `Data`. What do you think? Probably silly as this would be breaking. If I keep to `Collections` then no auto-updating needed. 

## Checklist:

### DIFF
--- a/Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Components/Collections/CreateCollection.cs
+++ b/Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Components/Collections/CreateCollection.cs
@@ -4,7 +4,6 @@ using Speckle.Connectors.GrasshopperShared.Components.BaseComponents;
 using Speckle.Connectors.GrasshopperShared.HostApp;
 using Speckle.Connectors.GrasshopperShared.Parameters;
 using Speckle.Connectors.GrasshopperShared.Properties;
-using Speckle.Sdk.Common;
 using Speckle.Sdk.Models.Collections;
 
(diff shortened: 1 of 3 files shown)


bimgeek (approved):
tested. looking good.

PR #1337, opened 2026-03-25
feat(revit): add shared location reference point setting

bjoernsteinhagen:
## Description

Fixes the `Survey` reference point option for send and receive workflows to correctly include True North rotation. 

Relying on `BuiltInParameter.BASEPOINT_ANGLETON_PARAM` frequently returned `0` in modern Revit versions. We now extract the angle natively via `document.ActiveProjectLocation.GetProjectPosition(XYZ.Zero).Angle`, which better grabs the correct rotation.

## User Value

Models exported from Revit using the Survey Point will now have the correct true north rotation applied.

## Changes:

- updated `ToSpeckleSettingsManager` and `ToHostSettingsManager` to handle the true north rotation for the `Survey` option using `ActiveProjectLocation.GetProjectPosition().Angle`

image

## Checklist:


bimgeek (approved):
tested, looking good.

PR #1344, opened 2026-03-27
feat: plant 3d connector

bimgeek:


Connectors/Autocad/Speckle.Connectors.AutocadShared/Plugin/AutocadCommand.cs
```diff
@@ -39,13 +42,23 @@ public void Command()
...
     // init DI
     var services = new ServiceCollection();
-    _disposableLogger = services.Initialize(AppUtils.App, AppUtils.Version);
+    try
+    {
+      _disposableLogger = services.Initialize(AppUtils.App, AppUtils.Version);
+    }
+    catch (BadImageFormatException)
+    {
+      // TODO: Speckle.Connectors.Logging assembly conflict in Plant3D — skip logging for now
```
JR-Morgan: What's the conflict? 
we have control over dependencies versions, and can align...

Converters/Plant3d/Speckle.Converters.Plant3dShared/Helpers/PropertyHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+  public bool TryGetValue<T>(Func<T> getValue, out T? value)
+  {
+    try
+    {
+      value = getValue();
+      return true;
+    }
+    catch (Exception e)
+      when (e is InvalidOperationException
+        || e is ArgumentException argEx && !argEx.IsFatal()
```
JR-Morgan, line 19: You don't need any of the `.IsFatal()` here... 
you're already filtering only for types that are going to return false
```cs
  [Pure]
  public static bool IsFatal(this Exception ex)
  {
    return ex switch
    {
      OutOfMemoryException
      or ThreadAbortException
      or InvalidProgramException
      or AccessViolationException
      or AppDomainUnloadedException
      or BadImageFormatException => true,
      _ => false,
    };
  }
```

Converters/Plant3d/Speckle.Converters.Plant3dShared/ToSpeckle/Geometry/Plant3dEntityToSpeckleConverter.cs
```diff
@@ -0,0 +1,121 @@
...
+        if (converter is not null)
+        {
+          var converted = converter.Convert(entity);
+          results.Add(converted);
+          return;
+        }
+      }
+      catch (System.Exception)
+      {
+        // Converter not found or failed — fall through to explode
```
JR-Morgan: this comment is a bit misleading, no?
converter not found will not reach this block right?

Converters/Plant3d/Speckle.Converters.Plant3dShared/ToSpeckle/Properties/Plant3dDataExtractor.cs
```diff
@@ -0,0 +1,88 @@
...
+          string key = kvp.Key?.ToString() ?? "unknown";
+          object? value = ToSerializable(kvp.Value);
+
+          result[key] = value;
+        }
+      }
+    }
+    catch (System.Exception)
+    {
+      // DataLinksManager not available or other API failure
```
JR-Morgan, line 60: Can we have a more explicit exception type here to avoid hiding unexpected exceptions?

Converters/Plant3d/Speckle.Converters.Plant3dShared/ToSpeckle/Properties/Plant3dDataExtractor.cs
```diff
@@ -0,0 +1,88 @@
...
+    {
+      null => null,
+      string s => s,
+      bool b => b,
+      int i => i,
+      long l => l,
+      double d => d,
+      float f => f,
+      decimal m => (double)m,
+      DateTime dt => dt.ToString("o"),
```
JR-Morgan: Speckle's serializer supports `decimal` and `DateTime` in a more optimal way that you've got here. I'd suggest letting them pass through.

And just FYI, there are other types we do support like enum, ushort, short, uint, ulong, Color, that are missing from this switch...  
but if Plant3d properties can't be of these types, it's not going to be useful to add them.

JR-Morgan (changes requested):

Converters/Plant3d/Speckle.Converters.Plant3dShared/Plant3dRootToSpeckleConverter.cs
```diff
@@ -0,0 +1,98 @@
...
+          }
+
+          tr.Commit();
+          return result;
+        }
+      }
+    }
+    catch (SpeckleException e)
+    {
+      Console.WriteLine(e);
```
JR-Morgan: Please can we use a proper logger (log info or log warning)
or simply not log at all since we're throwing anyway....

Converters/Plant3d/Speckle.Converters.Plant3dShared/Plant3dRootToSpeckleConverter.cs
```diff
@@ -0,0 +1,98 @@
...
+      );
+    }
+
+    Type type = dbObject.GetType();
+
+    var objectConverter = _toSpeckle.ResolveConverter(type);
+
+    try
+    {
+      using (var l = _settingsStore.Current.Document.LockDocument())
```
JR-Morgan: nitpick, but in general the using *declaration* is preferred over the using *statement*, unless you want more explicit control over the scope.

```cs
using var l = _settingsStore.Current.Document.LockDocument();
```

vs

```
using (var l = _settingsStore.Current.Document.LockDocument())
{
  ...
}
```

Converters/Plant3d/Speckle.Converters.Plant3dShared/ServiceRegistration.cs
```diff
@@ -0,0 +1,44 @@
+using Microsoft.Extensions.DependencyInjection;
+using Speckle.Converters.Autocad;
+using Speckle.Converters.Autocad.ToHost.Helpers;
+using Speckle.Converters.Common;
+using Speckle.Converters.Common.Registration;
+// using Speckle.Converters.Plant3dShared.Helpers;
```
JR-Morgan: remove if it's not needed

Sdk/Speckle.Connectors.Common/HostApplications.cs
```diff
@@ -16,6 +16,7 @@ public static class HostApplications
     Civil = new("Civil 3D", "civil3d"),
     Civil3D = new("Civil 3D", "civil3d"),
     AutoCAD = new("AutoCAD", "autocad"),
+    Plant3D = new("Plant3D", "plant3d"),
```
JR-Morgan: should be `Plant3D = new("Plant 3D", "plant3d"),`
with a space in "Plant 3D", to align with the official app's naming

JR-Morgan (changes requested):

Connectors/Autocad/Speckle.Connectors.AutocadShared/Plugin/AutocadCommand.cs
oguzhankoral: at early stage, i was having this and ignored for the sake of progress. can you get rid of and test? it might not valid anymore

Converters/Plant3d/Speckle.Converters.Plant3dShared/ToSpeckle/Geometry/Plant3dEntityToSpeckleConverter.cs
JR-Morgan: Actually, the comment is correct, because resolving the converter will throw a `ConversionNotSupportedException`.
It will however never return null, so the null check above is misleading.

I've removed the unecessary null check, and clarified the comment some more for good measure.

Converters/Plant3d/Speckle.Converters.Plant3dShared/ToSpeckle/Properties/Plant3dDataExtractor.cs
JR-Morgan: Actually, it looks like the properties are always strings. It's typed as `Dictionary<string, string>`, so I'm a bit confused why this switch statement was added to begin with.

I've removed it, but will ask @oguzhankoral for clarification in case there's something I'm not seeing....

JR-Morgan (dismissed):

JR-Morgan (dismissed):

dogukankaratas (approved):

JR-Morgan (approved):

PR #1345, opened 2026-03-30
fix(rhino): resolve material type editing

bjoernsteinhagen:
## Description

Fixes a bug where materials created during a load operation couldn't have their types changed (e.g., from Custom to Glass) without duplicating the material. 

The core issue was that we were populating the legacy Material table directly. When changing type, UI switches to `RenderMaterial` which is why we get a duplicate. So, when we only provided legacy materials, Rhino created "Custom" wrappers that broke when edited. By switching to creating `RenderMaterial` directly, we fix this.

During this change, I also need to adopt `Guid` for material tracking. This bypasses the race conditions where the legacy table hasn't synced with the RDK yet, which was causing the "Default Material" assignment bug on received objects.

Fixes [CNX-2896](https://linear.app/speckle/issue/CNX-2896/cant-change-material-type-of-created-material-by-speckle).

## User Value

Users can now natively edit Speckle-received materials in the Rhino Materials panel and change their types (Glass, Plaster, etc.) without breaking the model or creating duplicate materials.

## Changes:

- **RhinoMaterialBaker**: creates `RenderMaterial` wrappers instead of just legacy entries.
- **POC Change**: Replaced integer index mapping with `Guid` mapping (`ObjectIdAndMaterialIdMap`)
- **Bakers Update**: Updated `RhinoLayerBaker`, `RhinoInstanceBaker`, and `RhinoHostObjectBuilder` to assign materials via `RenderMaterialId`.

### Before
- Load model
- Select object
- Edit type of associated material
- Immediately material duplicated ❌ 

https://github.com/user-attachments/assets/0875f7ac-4da6-4574-a5e8-666aa23032ac

### After
- Load model
- Select object
- Edit type of associated material. Edit done in place. ✅ 

https://github.com/user-attachments/assets/4dc35268-5212-46d6-8b25-0b0b6e76d1e6

## Checklist:

### DIFF
--- a/Connectors/Rhino/Speckle.Connectors.RhinoShared/HostApp/RhinoInstanceBaker.cs
+++ b/Connectors/Rhino/Speckle.Connectors.RhinoShared/HostApp/RhinoInstanceBaker.cs
@@ -2,6 +2,7 @@ using Microsoft.Extensions.Logging;
 using Rhino;
 using Rhino.DocObjects;
 using Rhino.Geometry;
+using Rhino.Render;
 using Speckle.Connectors.Common.Conversion;
 using Speckle.Connectors.Common.Instances;
 using Speckle.Connectors.Common.Operations;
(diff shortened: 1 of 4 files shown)


bimgeek (approved):
tested and can confirm the issue has been resolved.

oguzhankoral (approved):

PR #1349, opened 2026-04-01
feat(connectors): disable cache config

bjoernsteinhagen:
## Description

Wires up the backend to `DisableCache` setting sent from the DUI. Publish intercepted at `SendOperationManager` level and clears the `ISendConversionCache`. 

Addresses [CNX-3261](https://linear.app/speckle/issue/CNX-3261/add-disable-cache-setting-to-prevent-stale-parameter-updates-on).

## User Value

See [CNX-3261](https://linear.app/speckle/issue/CNX-3261/add-disable-cache-setting-to-prevent-stale-parameter-updates-on).

## Changes:

- Added `DisableCache` property to the `ConnectorConfig` in `ConfigStore.cs` (defaults to `false`).
- Injected cache clearing logic into `SendOperationManager.cs` right before the `gatherObjects` execution.

## To-do before merge:

- [ ] blocked by #1350 

## Screenshots & Validation of Changes:

https://github.com/user-attachments/assets/6003ad63-0153-42b2-8bd8-96c9bf3208e8

## Checklist:


JR-Morgan (commented):
I'm guessing we'll still be doing all the change tracking (host app API event listening) to invalidate the cache 
even tho the cache is disabled

DUI3/Speckle.Connectors.DUI/Bindings/SendOperationManager.cs
```diff
@@ -92,6 +90,14 @@ public async Task Process<T>(
 
       initializeScope(serviceScope.ServiceProvider, modelCard);
 
+      // if user has disabled cache, wipe the in-memory cache before gathering and building objects
```
oguzhankoral, line 96: clear the cache if any item exists on the cache for sure
but also, prevent adding object references to cache when the cache setting is disabled

oguzhankoral (changes requested):

DUI3/Speckle.Connectors.DUI/Bindings/SendOperationManager.cs
bjoernsteinhagen: [feat(connectors): prevents StoreSendResult if cache disabled](https://github.com/specklesystems/speckle-sharp-connectors/pull/1349/commits/890bc2c4e826b4c04a9d69bd823505ec791a1410)

oguzhankoral:
> I'm guessing we'll still be doing all the change tracking (host app API event listening) to invalidate the cache even tho the cache is disabled

@JR-Morgan's point is legit. i guess we were doing some subs/unsubs on events

DUI3/Speckle.Connectors.DUI/Bindings/SendOperationManager.cs
```diff
@@ -95,7 +95,8 @@ public async Task Process<T>(
       if (configStore.GetConnectorConfig().DisableCache)
       {
         var sendConversionCache = serviceScope.ServiceProvider.GetRequiredService<ISendConversionCache>();
-        sendConversionCache.ClearCache();
+        sendConversionCache.ClearCache(); // clear whatever is currently in there to ensure 0% cache hits
+        sendConversionCache.IsBypassed = true; // tells cache to ignore any future write requests during this scoped operation
```
oguzhankoral: who makes it `false` when cache enabled?

oguzhankoral (changes requested):

DUI3/Speckle.Connectors.DUI/Bindings/SendOperationManager.cs
bjoernsteinhagen: [fix(connectors): restores IsBypassed state](https://github.com/specklesystems/speckle-sharp-connectors/pull/1349/commits/b60797628c5809cbcfcf6aa493680fda6e7c97dc). thanks for that! Added in finally block so we always restore initial state (regardless of failed operations, cancelled operations etc.)

bjoernsteinhagen:
> I'm guessing we'll still be doing all the change tracking (host app API event listening) to invalidate the cache even tho the cache is disabled

@JR-Morgan I'm a bit uneasy about my understanding / changes. So made a PR in to this working branch here #1350.

DUI3/Speckle.Connectors.DUI/Bindings/SendOperationManager.cs
oguzhankoral: deep in me still not happy with this flag. i will try to implement explicit flag to pass relavant function instead setting here silently and reverting back somewhere else.

JR-Morgan (dismissed):

JR-Morgan (approved):

oguzhankoral (approved):

PR #1361, opened 2026-04-09
fix(revit): Lighting source material is included in the material quantities

bimgeek:
`GetMaterialIds()` on lighting fixture elements returns the IDs of all materials applied to that element, including materials applied to lighting cone geometry (the visual "light source" cone/sphere Revit renders). This causes spurious entries in `properties["Material Quantities"]` for lighting fixtures, e.g. a material named "Default Light Source" with a non-zero area/volume.

The display-value side of the pipeline already excludes this geometry in `DisplayValueExtractor.SkipGeometry()` by checking `OST_LightingFixtureSource` on the geometry object's graphics style. However, the Revit API offers no equivalent on `GetMaterialIds()`. There is no way to ask "which material IDs belong to light source geometry vs. structural geometry". The fix is to exclude by the well-known default name.

**Known limitation:** users who have renamed or replaced the default light source material with a custom name will not be covered.

### DIFF
--- a/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/Raw/MaterialQuantitiesToSpeckle.cs
+++ b/Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/Raw/MaterialQuantitiesToSpeckle.cs
@@ -32,6 +32,14 @@ public class MaterialQuantitiesToSpeckleLite : ITypedConverter<DB.Element, Dicti
   private readonly IConverterSettingsStore<RevitConversionSettings> _converterSettings;
   private readonly StructuralMaterialAssetExtractor _structuralAssetExtractor;
 
+  // Materials excluded from quantity extraction. Revit's GetMaterialIds() returns materials
+  // applied to geometry that we already exclude from display values (e.g. lighting cone).
+  // The API provides no way to identify such materials programmatically, so we exclude by name.
+  // Known limitation: custom material names are not covered.
+  private static readonly HashSet<string> s_excludedMaterialNames = ["Default Light Source"];
+
+  private static bool IsExcludedMaterial(string materialName) => s_excludedMaterialNames.Contains(materialName);
+
   public MaterialQuantitiesToSpeckleLite(
     ScalingServiceToSpeckle scalingService,
     IConverterSettingsStore<RevitConversionSettings> converterSettings,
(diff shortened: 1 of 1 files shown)


Converters/Revit/Speckle.Converters.RevitShared/ToSpeckle/Raw/MaterialQuantitiesToSpeckle.cs
```diff
@@ -203,6 +203,11 @@ out string matName
     matName = "";
     if (_converterSettings.Current.Document.GetElement(matId) is DB.Material material)
     {
+      // No API to identify light-cone materials by ID; exclude by well-known default name.
+      if (material.Name == "Default Light Source")
```
bjoernsteinhagen, line 207: It's a pity about the lack of an ID because this will only work for English, then?

bimgeek: yep, that's the way it is. I think we can be flexible here. If people ask for it, we can expand this list to cover other languages too.

bjoernsteinhagen: Assuming that `material.Name` is the same across all supported revit versions and that this has been tested and works?

bimgeek: `Material.Name` is inherited from `Element` class. it's a foundational property on the base class that every Revit element exposes.

bjoernsteinhagen: I meant the string of "Default Light Source" for `material.Name` ... For all rvt versions that exact string gets used?

bimgeek: Oh, you probably mean if it's "Default Light Source"... Yes, that is consistent AFAIK. I'll check with Revit 2022 just in case, which is the earliest version we support.

bjoernsteinhagen (approved):
also double-checked, unfortunately no dedicated api mechanism to identify and differentiate light cone materials (and therefore volumes) apart from the string approach. lgtm

PR #1367, opened 2026-04-16
fix(rhino-importer): Fix rhino retrying already failed jobs 

JR-Morgan:
see https://linear.app/speckle/issue/CXPLA-482/investigate-rhino-importer-incorrect-re-queueing-behaviour

live and tested on latest + testing 3


Importers/Rhino/Speckle.Importers.JobProcessor/JobProcessor.cs
```diff
@@ -170,6 +194,7 @@ private async Task AttemptJob(FileimportJob job, IDbConnection connection, Cance
 
       await ExecuteJobWithTimeout(job, speckleClient, serviceCancellationToken);
       totalElapsedSeconds = stopwatch.Elapsed.TotalSeconds;
+      await repository.FinishJob(connection, job.Id, CancellationToken.None);
```
JR-Morgan, line 192: Main change no. 1, jobs are marked as successful as soon as the importer has completed them.

Importers/Rhino/Speckle.Importers.JobProcessor/JobProcessor.cs
```diff
@@ -118,14 +125,31 @@ await client.Ingestion.FailWithCancel(
...
   private async Task ReportFailed(
+    IDbConnection connection,
     FileimportJob job,
     IClient client,
     Exception ex,
     double elapsedSeconds,
     CancellationToken cancellationToken
   )
   {
+    await repository.FailJob(connection, job.Id, cancellationToken);
```
JR-Morgan, line 146: Change no. 2, marking jobs failed for all "failed paths" (including cancelled)

Importers/Rhino/Speckle.Importers.JobProcessor/JobProcessor.cs
```diff
@@ -188,38 +213,27 @@ private async Task AttemptJob(FileimportJob job, IDbConnection connection, Cance
...
         logger.LogError(new AggregateException(ex, ex2), "Failed to report failure status");
-        // somehow we're in a weird state,
-        // let's return the job to the queued state where it will get picked up again until one of total timeout,
-        // max attempts, or exhausted compute budget is reached.
-        // The server is responsible for garbage collecting jobs which have reached these error conditions and moving
-        // them to a failed status.
-        await repository.ReturnJobToQueued(connection, job.Id, CancellationToken.None);
+        // somehow we're in a weird state, e.g. we couldn't report the ingestion failure
+        // The server will clean up the ingestion if it still thinks it's processing.
+        await repository.FailJob(connection, job.Id, CancellationToken.None);
```
JR-Morgan, line 236: Change no. 3, don't requeue when we can't report ingestion faliure. Lets go straight to failed.

Not too sure about this change. I can't remember why we prefered the current state

JR-Morgan: I suspect the reason for the old behaviour is no longer valid now ingestions have their own separate clean up mechanism. And we don't `file_uploads` table to keep in sync with the `background_jobs` table

JR-Morgan:
currently live on latest + testing environments

gjedlicska (approved):

PR #1370, opened 2026-04-20
feat!: STJ Experiment

JR-Morgan:
see https://github.com/specklesystems/speckle-sharp-sdk/pull/468 for the SDK changes.

This PR updates this repo to use the new System.Text.Json (STJ) powered SDK.

See https://www.notion.so/speckle/Introducing-System-Text-Json-to-packfile-based-serializer-SendPipleine-357b78fc7aa6805bb472ca09b980b7fd for context.

---

## PR Round 2
The first version of this PR bumped the minimum version of the SDK to 8.0.0. This was  bad for several reasons.
(compatibility with older Revit plugins, and potentially dangerous situations in Rhino .NET7/4.8)

This this new version of the PR is far more safe. Our min compatible STJ version is now 5.0.0; (same as what v2 connectors used). No longer conflicts with any other plugins.
Downsides are pretty minimal, it does limit our SDK to 5.0.0 API, but this is workable.

For .NET 8/10 targets
 - No change dependency from `main`, we're good.

For .NET4.8 (including Rhino8): 
   - Plugins that that are compatible with any version  ≤8.0.6 STJ will not be broken by speckle
   - Plugins that load STJ 5.0.0 (or greater) guaranteed to not break Speckle.
   - tl;dr very little opportunity for dll conflicts


JR-Morgan:
We should also test Grasshopper, and Rhino under .net7 (rather than .net8)

JR-Morgan:
~~Pulling my hair out a bit, I thought that dropping our min verison to 5.0.0 would be the end of any conflicts. But it looks like the .NET Standard 2.0 STJ dll is not compatible with .NET 4.6.2 STJ dll.~~

~~This is very strange, so needs more investigation.
as it is right now, we still have a conflict with Data Exchange, because we require .NET Standard2.0 dll, and it requires .NET 4.6.2. Normally these would be compatible, but for what ever reason, they are not 😢.
This is probably a bug in STJ that was never fixed in 5.x~~

~~Adding a .net 4.6.2 target to our SDK would probably solve this, but that would need a discussion.~~

~~A better solution might be to use some reflection to work around this. I'll try `Activator.CreateInstance` to see if that's more forgiving.~~

Fixed by downgrading System.Memory from 4.5.5 -> 4.5.3

JR-Morgan:
Current status:
we've released this as a beta (see https://speckle.community/t/connector-performance-improvements-beta/22277)

no reports of problems currently, but we haven't seen that many downloads yet. so I want to give this longer to simmer.

Sdk/Speckle.Connectors.Common/AssemblyCompatibilityCheck.cs
```diff
@@ -0,0 +1,95 @@
...
+  /// under runtime conditions where STJ binds to a different version of System.Buffers than our SDK
+  /// </summary>
+  /// <remarks>
+  /// We're being extra careful introducing System.Text.Json SDK changes in connectors.
+  /// This check will alert us to any conflicts... and help us avoid breaking sends.
+  /// We're only really concerned with this being a possibility in .NET Framework apps that don't bundle STJ themselves.
+  /// If we wish, we can remove this check at a later date if we're confident there's no issues.
+  /// </remarks>
+  /// <returns></returns>
+  public bool ValidateStjCompatiblity()
```
bjoernsteinhagen: typo with `Compatibility`

Sdk/Speckle.Connectors.Common/Operations/SendOperation.cs
```diff
@@ -54,7 +55,10 @@ CancellationToken cancellationToken
     if (useModelIngestionSend)
     {
       bool usePackfileSend =
-        rootContinuousTraversalBuilder != null && await CheckPackfileSendEndpoints(sendInfo, cancellationToken);
+        rootContinuousTraversalBuilder != null
+        && await CheckPackfileSendEndpoints(sendInfo, cancellationToken)
+        && compatabilityCheck.ValidateStjCompatiblity();
```
bjoernsteinhagen: Is `ValidateStjCompatiblity()` called on every send? If so, should the result be cached — especially if two sends run concurrently? Or perf. hit negligible?

JR-Morgan: Because assemblies are loaded lazily (as and when they are used), it's possible for another plugin to load an extra dll in the time between sends.
So It's a good idea to check at the start of every send.

It also doesn't take very long to validate. It's just serializing a small tiny string and observing an exception.

As for concurrent sends, I don't think it will have any negative affect.

Sdk/Speckle.Connectors.Common/AssemblyCompatibilityCheck.cs
JR-Morgan: fixed, good catch 😅

bjoernsteinhagen (approved):

PR #1371, opened 2026-04-20
feat (rhino): dataobject wrapper

dogukankaratas:
Wraps rhino geometry in the new RhinoObject type so data sent from Rhino matches with general connector extracted data.

Send:
- `RhinoRootObjectBuilder.ConvertRhinoObject` and `RhinoContinuousTraversalBuilder.ConvertRhinoObject` now build a RhinoObject with:
    - displayValue = mesh prims only (for viewer)
    - rawEncoding = native 3dm blob for Breps/Extrusions/SubDs (for receive roundtrip)
    - name, type, units, properties as typed fields

Receive:
- DataObjectConverter now checks for rawEncoding on incoming RhinoObject and decodes it to native Brep/Extrusion/SubD. Falls back to displayValue meshes only when no rawEncoding is present.

Instance handling has done with the following approach:

image

Related SDK changes: https://github.com/specklesystems/speckle-sharp-sdk/pull/473

### DIFF
--- a/Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Operations/Receive/GrasshopperBlockUnpacker.cs
+++ b/Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Operations/Receive/GrasshopperBlockUnpacker.cs
@@ -1,3 +1,4 @@
+using Microsoft.Extensions.Logging;
 using Rhino.Geometry;
 using Speckle.Connectors.Common.Operations.Receive;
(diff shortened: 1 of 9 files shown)


bimgeek (approved):
testing is good. Sent all geometry types, properties, materials, blocks...
Received in other host apps (Rhino, GH, Revit, SketchUp, Blender etc.) all good.

JR-Morgan (approved):

PR #1374, opened 2026-04-21
fix(grasshopper): fix instance proxies now with rhino data object wra…

bjoernsteinhagen:
## Description

When receiving a model from rhino in grasshopper, block instances were appearing at the origin with no transforms applied. This is because the recent rhino changes now wrap every atomic object (including block definition members) in a `RhinoDataObject`. The grasshopper receive handler was hitting the DataObject branch for these, which never populated `ConvertedObjectsMap`. So when the block unpacker later tried to resolve definition members by id, it found nothing. Definitions built as null, instances got dropped silently, and the def geometry rendered as standalone objects on their original layers — which looked like "transforms aren't applying."

## Changes:

- `LocalToGlobalMapHandler.ConvertObjectToCache`: populate `ConvertedObjectsMap` from the DataObject branch, keyed by the DataObject's applicationId
- `GrasshopperCollectionRebuilder.RemoveConsumedObjectsFromCollection`: extended to also match `SpeckleDataObjectWrapper` entries, so def members get stripped from their original layer after block unpacking.

## Validation of changes:

TBD


dogukankaratas (approved):
tested and looks good

PR #1379, opened 2026-04-23
feat (acad, rhino): dataobject wrappers

dogukankaratas:
Wraps primitive geometries into DataObjects for AutoCAD and Rhino (Civil3D was sending mix-typed objects, this PR is also unifies the send data using only DataObjects).

Here is the hierarchy behind it:

image

Builds a RhinoObject or AutocadObject with: 
- displayValue: mesh prims only. 
- rawEncoding (for breps and solids): native 3dm or SAT blob for receive roundtrip.
- name, type, units and properties. 

Also adds 'Attributes' field to the Autocad Blocks.

For receive side:
- DataObjectConverter now checks for rawEncoding on incoming RhinoObject and decodes it to native Brep/Extrusion/SubD. Falls back to displayValue meshes only when no rawEncoding is present.
- Aligns GH and Revit for a proper geometry interop.


Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadInstanceUnpacker.cs
```diff
@@ -78,7 +78,13 @@ private void UnpackInstance(BlockReference instance, int depth, Transaction tran
...
 
       var properties = _propertiesExtractor.GetProperties(instance);
-      if (properties?.Count > 0)
+      var attributes = GetInstanceAttributes(instance, transaction);
+      if (attributes.Count > 0)
+      {
+        properties["Attributes"] = attributes;
+      }
+
+      if (properties.Count > 0)
```
bjoernsteinhagen, line 87: previous code had null-conditional on properties. could properties  be null here? if it can be null and attributes exist, this would blow up?

Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadInstanceUnpacker.cs
```diff
@@ -185,4 +191,17 @@ out InstanceDefinitionProxy? value
...
+    var attributes = new Dictionary<string, object?>();
+
+    foreach (ObjectId id in instance.AttributeCollection)
+    {
+      var reference = (AttributeReference)transaction.GetObject(id, OpenMode.ForRead);
+      attributes[reference.Tag] = reference.TextString;
+    }
+
+    return attributes;
+  }
```
bjoernsteinhagen: so `GetInstanceAttributes` iterates `instance.AttributeCollection` and casts each `ObjectId` to `AttributeReference`? no try/catch? i see most autocad wraps caller's in a try/catch or have its own guard. is this 100 % safe?

bjoernsteinhagen (changes requested):

Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadInstanceUnpacker.cs
dogukankaratas: Fair. Wrapped each attribute in a try catch.

bjoernsteinhagen (approved):

PR #1381, opened 2026-04-27
fix (plant3d): add labels

dogukankaratas:
### Problem
AutoCAD Plant 3D ATTDEF (attribute definition) objects are not being correctly extracted by the connector. When published to Speckle, the object appears in the selection tree but has no visual representation, and its tag value is sent as an empty string. This means any text defined via the ATTDEF command is silently lost on publish.

### Solution
`AttributeDefinition` inherits from `DBText`. Without a converter it falls back to `DBTextToSpeckleConverter` but the default value for `TextString` is empty for this. Adding a new converter for attribute definition type maps the conversion to the correct target.

Here is the attribute definition in Plant3D
image

Before the implementation it was empty visually
image

Now it converts correctly. 
image

Also previously, CollectDisplayObjects() exploded BlockReferences and converted AttributeDefinitions, which exposed template strings. The converter now reads visible AttributeReferences from the instance AttributeCollection, converts them through the existing DBText path.

How it looks in Plant3D:
image

Before:
image

Now:
image


bimgeek (approved):
tested, looking good.

PR #1394, opened 2026-05-09
fix(grasshopper): IFC DataObjects lose geometry and inflate object count on republish

bjoernsteinhagen:
## Description

Two bugs affecting the IFC → Grasshopper → republish → re-receive workflow.

**Bug 1: DataObjects lose geometry on re-receive**
IFC meshes use `@`-prefix chunked arrays (e.g. `@vertices_5000_0`) that don't map back to typed properties on deserialisation. Publishing `wrapper.Base` as-is meant `mesh.vertices == null` on re-receive, crashing the mesh converter. The crash was silently swallowed, dropping the entire DataObject. Fix: reconvert from `wrapper.GeometryBase` (the live Rhino mesh) to produce a clean Speckle object before publishing.

**Bug 2: QuerySpeckleObjects returns inflated object count on re-receive**
IFC DataObjects carry structural dynamic properties (e.g. an `elements` sub-tree representing the IFC hierarchy). Publishing `wrapper.DataObject` as-is meant GraphTraversal recursed into `elements` on re-receive, inflating counts (e.g. 68 published → 5961 received). Fix: build a fresh `DataObject` containing only `applicationId`, `name`, `properties`, and `displayValue`.

**Refactor: extracted shared logic to `GrasshopperSendUnwrapper`**
`UnwrapGeometry` and `UnwrapDataObject` were identical across `GrasshopperRootObjectBuilder` and `GrasshopperContinuousTraversalBuilder`. Extracted to a shared static helper — both bugs needed fixing in both places, which is what surfaced the duplication.

## User Value

IFC DataObjects published from Grasshopper now round-trip correctly — geometry is preserved and the received object count matches what was published.

## Changes:

- Added `GrasshopperSendUnwrapper` — shared static helper with `UnwrapGeometry` and `UnwrapDataObject`
- `GrasshopperRootObjectBuilder` and `GrasshopperContinuousTraversalBuilder` now delegate to the helper
- `UnwrapGeometry`: when Rhino geometry is available, reconverts to a clean Speckle object instead of falling back to the original IFC `Base`
- `UnwrapDataObject`: builds a fresh `DataObject` instead of mutating the original, stripping IFC structural dynamic properties

### Reported issue

1. Load large IFC model (31134 `DataObject` with `displayValue`, 48 `DataObject` without `displayValue`)

image

2. Randomly select 20 `DataObject` with `displayValue`. 

image

3. Have those `DataObject` pass through a couple components (including `SpeckleDataObjectPassthrough` and `Merge`)
4. Merge back with 48 `DataObject` without `displayValue`
5. Publish 68 `DataObject`

image

6. Re-recevie in GH and ensure we receive back 20 `DataObject` with `displayValue` and 48 `DataObject` without

**Before:** 68 DataObjects published
- 5961 objects on re-receive ❌ 
- 7/20 DOs with geometry ❌ 

image

**After:** 68 DataObjects published
- 68 objects on re-receive ✅
- 20/20 DOs with geometry ✅ 

image

### Regression check
- Load large revit model
- Query speckle objects to get object count
- Re-publish
- Load again and verify objects count

image

## Checklist:


Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Operations/Send/GrasshopperSendUnwrapper.cs
```diff
@@ -0,0 +1,83 @@
...
+  /// Saw some IFC objects arrive with @-prefix chunked arrays (e.g. <c>@vertices_5000_0</c>) that don't map
+  /// to typed properties on deserialization — publishing them as-is crashes the mesh converter on re-receive
+  /// (See CNX-3303).
+  /// </remarks>
+  public static Base UnwrapGeometry(SpeckleGeometryWrapper wrapper)
+  {
+    Dictionary<string, object?> props = [];
+    Base baseObject = wrapper.Base;
+
+    if (wrapper.GeometryBase != null)
```
dogukankaratas: Could this break block instances? SpeckleBlockInstanceWrapper also inherits SpeckleGeometryWrapper and always has a GeometryBase, so it would hit this reconvert path too.

bjoernsteinhagen: good catch. fix in [fix(grashopper): guard block instances from geometry reconversion path](https://github.com/specklesystems/speckle-sharp-connectors/pull/1394/commits/1995ff3d23b679fd539ae349b7871d5df5c7ded4)

dogukankaratas (approved):
lgtm

PR #1395, opened 2026-05-12
feat (civil3d): parameter updater

dogukankaratas:
Implements parameter updater functionality in Civil3D.


Connectors/Autocad/Speckle.Connectors.Civil3dShared/HostApp/PropertyUpdater.cs
```diff
@@ -0,0 +1,174 @@
...
+          {
+            continue;
+          }
+
+          var propDef = FindPropertyDefinition(propertySet, setDefinition, targetPropName, internalDefinitionName);
+          if (propDef is not null)
+          {
+            propertySet.UpgradeOpen();
+            propertySet.SetAt(propDef.Id, CoerceValue(newValue, propDef.DataType));
+            return UpdateResult.Success();
```
bjoernsteinhagen, line 63: Can `SetAt` throw? 

Also, `PropertySetBaker` uses a different pattern which seems simpler? (using `ForWrite`?)

Connectors/Autocad/Speckle.Connectors.Civil3dShared/Bindings/Civil3dParametersBinding.cs
```diff
@@ -0,0 +1,222 @@
...
+    if (rawPath.StartsWith("properties.", StringComparison.InvariantCultureIgnoreCase))
+    {
+      rawPath = rawPath[11..];
+    }
+    else if (rawPath.StartsWith("parameters.", StringComparison.InvariantCultureIgnoreCase))
+    {
+      rawPath = rawPath[11..];
+    }
+
+    var pathParts = rawPath.Split(['.'], 3);
```
bjoernsteinhagen, line 212: can C3D PropertySet names include dots?

bjoernsteinhagen (changes requested):

bjoernsteinhagen (changes requested):
`ParameterChangeRequest`, `ParameterChangesWrapper` and `UpdateResult` are identical between Rvt and C3D with what seems like no host app dependency -> move to DUI namespace? There is even a TODO in `UpdateResult` about this ...

Connectors/Autocad/Speckle.Connectors.Civil3dShared/HostApp/PropertyUpdater.cs
dogukankaratas: yes, `SetAt` can throw. the update block is wrapped in a catch, so in that case we log it and return an `UpdateResult.Fail` instead of failing the whole batch.

`ForWrite` would also work, but here I open the `PropertySet` for read while scanning and only upgrade the matched one before writing, to avoid opening every candidate `PropertySet` for write.

Connectors/Autocad/Speckle.Connectors.Civil3dShared/Bindings/Civil3dParametersBinding.cs
dogukankaratas: I’d rather not add host-side guessing here; if we need to support that case, we should make the payload more explicit instead of relying on path parsing.

dogukankaratas:
> `ParameterChangeRequest`, `ParameterChangesWrapper` and `UpdateResult` are identical between Rvt and C3D with what seems like no host app dependency -> move to DUI namespace? There is even a TODO in `UpdateResult` about this ...



> `ParameterChangeRequest`, `ParameterChangesWrapper` and `UpdateResult` are identical between Rvt and C3D with what seems like no host app dependency -> move to DUI namespace? There is even a TODO in `UpdateResult` about this ...

Good call. Applied.

Connectors/Autocad/Speckle.Connectors.Civil3dShared/HostApp/PropertyUpdater.cs
bjoernsteinhagen: makes sense sir!

bjoernsteinhagen (approved):
nice

PR #1399, opened 2026-05-14
feat(navis)!: Add Navis 2027

JR-Morgan:


bimgeek (approved):
tested. Navis 2027 is good.

bimgeek (approved):

PR #1404, opened 2026-05-17
feat(navisworks): performance optimisations — geometry welding, property detail levels, and quick property caching

jsdbroughton:
## Summary

A performance-focused overhaul of the Navisworks connector's send pipeline, targeting large model conversion times through geometry welding, property extraction granularity controls, and a new quick property definitions cache.

### Geometry
- **Geometry Detail Level** setting (new): `Full`, `Optimised` (default), or `Lite`
-   - `Optimised` applies 65-degree seam-aware vertex welding and unit-aware coordinate rounding — this is the new default
-   - `Full` sends raw geometry with no post-processing (previous implicit behaviour)
-   - `Lite` skips geometry entirely (properties/hierarchy only)
- - Bounding box conversion added as a lightweight geometry fallback
- - 2D element path management and correct material assignment for non-solid geometry
- - **`RoundMeshVertexDoubles`** exposed as a standalone QA/QC toggle (unit-aware rounding without full optimisation)
### Properties
- **Property Detail Level** setting (new): `All`, `Standard` (default), or `None`
-   - `All`: full user-filtered property set
-   - `Standard`: filtered properties, minus the exclusion list, unless in Quick Properties — this is the new default
-   - `None`: item identity only (fastest)
- - `ExcludePropertiesSetting` introduced; legacy `excludeProperties: true` cards map to `None` for backwards compatibility
- - Internal Navisworks properties (formerly never extracted) can be included via `IncludeInternalProperties` (hidden, defaults `false`)
- - Conditional property merging in the Speckle converter based on detail level
### Performance infrastructure
- `QuickPropertyDefinitionsCache` — file-scope cache of property category/name pairs, built once and reused across all elements in a send operation
- - `QuickPropertyDefinition`, `QuickPropertyDefinitionsUtil`, and `QuickPropertyMatching` utilities replace repeated Navisworks API traversal
- - Property handler refactored to pull from the local settings store rather than constructor injection
- - Property extraction metrics tracking added
### Packages / housekeeping
- `System.Text.Json`, `System.Buffers`, `Microsoft.Bcl.AsyncInterfaces` updated to latest versions
- - Regex replaced with string operations in hot paths
- - BOM removed from affected files; import ordering standardised
## Default behaviour changes

| Behaviour | Before this PR | After this PR |
|---|---|---|
| Geometry processing | Raw triangles, no welding or rounding | **Optimised** by default: 65° seam-aware welding + unit-aware vertex rounding |
| Property extraction | All user-visible properties always sent | **Standard** by default: filtered minus exclusion list, Quick Properties always included |
| Internal Navisworks properties | Never sent | Still not sent by default (`IncludeInternalProperties = false`); opt-in available |
| Geometry skip option | Not available | **Lite** mode skips geometry entirely |
| Legacy `excludeProperties: true` cards | Behaviour undefined post-refactor | Mapped to `PropertyDetailLevel.None` for backwards compatibility |

The key user-facing change is that a default send will now produce **smaller, cleaner meshes** (welded, rounded to unit precision) and a **reduced property payload** (Standard level rather than All). Both can be overridden per model card in the UI settings.


Connectors/Navisworks/Speckle.Connectors.NavisworksShared/Operations/Send/Settings/GeometryDetailLevelSetting.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Speckle.Connector.Navisworks.Operations.Send.Settings;
+
+public class GeometryDetailLevelSetting(GeometryDetailLevel value) : ICardSetting
+{
+  public string? Id { get; set; } = "geometryDetailLevel";
+  public string? Title { get; set; } = "Geometry Detail";
+  public string? Description { get; set; } = "Full, Optimised (65 degree seam-aware + unit-aware rounding), or Lite.";
+  public string? Type { get; set; } = "string";
+  public List<string>? Enum { get; set; } =
```
JR-Morgan, line 16: What will the default value for this be? @bjoernsteinhagen I remember settings defaults being a little unintuitive r.e. defaults

Converters/Navisworks/Speckle.Converters.NavisworksShared/Services/QuickPropertyMatching.cs
```diff
@@ -0,0 +1,103 @@
...
+      {
+        return true;
+      }
+    }
+
+    return false;
+  }
+
+  private static string SanitizePropertyName(string name) =>
+    name == "Item" ? "Item" : Regex.Replace(name, @"[\.\/\s]", "_");
```
JR-Morgan: If performance is a concern here, replacing the Regex with something like 
```
name.Replace('.', '_')
              .Replace('/', '_')
              .Replace(' ', '_');
```
would be faster (perhaps important if the function is used for every property name...)

JR-Morgan:
Hi @jsdbroughton, please could I get a summary on how the default behaviours have changed r.e. mesh optimisations.

The extra options available via settings are great, I just want to understand how the default behaviour has changed.

jsdbroughton:
> Hi @jsdbroughton, please could I get a summary on how the default behaviours have changed r.e. mesh optimisations.
> 
> The extra options available via settings are great, I just want to understand how the default behaviour has changed.

When i wrap up the draft PR into final I'll make sure this is all recorded in the description

Connectors/Navisworks/Speckle.Connectors.NavisworksShared/Operations/Send/Settings/GeometryDetailLevelSetting.cs
jsdbroughton: They were set in the constructor, but have modified both geometry settings and property settings to make the default explicit.

```csharp
 public const GeometryDetailLevel DEFAULT_VALUE = GeometryDetailLevel.Optimised;

  public string? Id { get; set; } = SETTING_ID;
  public string? Title { get; set; } = "Geometry Detail";
  public string? Description { get; set; } = "Full, Optimised (65 degree seam-aware + unit-aware rounding), or Lite.";
  public string? Type { get; set; } = "string";
  public List<string>? Enum { get; set; } =
  [GeometryDetailLevel.Full.ToString(), GeometryDetailLevel.Optimised.ToString(), GeometryDetailLevel.Lite.ToString()];
  public object? Value { get; set; } = value.ToString();
```

and 

```csharp
public class PropertyDetailLevelSetting(PropertyDetailLevel value = PropertyDetailLevelSetting.DEFAULT_VALUE)
  : ICardSetting
{
  public const string SETTING_ID = "propertyDetailLevel";
  public const PropertyDetailLevel DEFAULT_VALUE = PropertyDetailLevel.Standard;

  public string? Id { get; set; } = SETTING_ID;
  public string? Title { get; set; } = "Property Detail";
  public string? Description { get; set; } =
    "All: user-filtered properties. Standard: filtered minus exclusions unless in Quick Properties. None: Item only.";
  public string? Type { get; set; } = "string";
  public List<string>? Enum { get; set; } =
  [nameof(PropertyDetailLevel.All), nameof(PropertyDetailLevel.None), nameof(PropertyDetailLevel.Standard)];
  public object? Value { get; set; } = value.ToString();
```

jsdbroughton:
> When i wrap up the draft PR into final I'll make sure this is all recorded in the description

Done

bimgeek (approved):

PR #1406, opened 2026-05-18
fix(grasshopper): collections reject bare Speckle Geometry as direct input

bjoernsteinhagen:
## Description

Collections were accepting bare Speckle Geometry as direct elements, breaking downstream workflows that expect Data Objects only.

Fixes [#ENG-8234](https://linear.app/speckle/issue/ENG-8234/gh-to-drop-speckle-geometry-component)

## User Value

Collections now enforce that only Data Objects and sub-collections can be piped in — geometry gets a clear error message telling you to wrap it in a Speckle Data Object first.

## To-do before merge
- [x] @bimgeek please test on QA/QC side including backwards compatibility (load "old" model that has collections with `Speckle Geometry` in it)

## Changes:

- `CreateCollection` rejects bare geometry inputs inline with an error + early return
- `SendComponent` and `SendAsyncComponent` apply the same rule in their mixed-input path
- `CollectionHelpers.HasAnyValidContent` retains `SpeckleGeometryWrapper` support (legacy receive path — models published before ENG-8234 may still have bare geometry in collections)

## Validation of changes:

image

## Checklist:


bimgeek (approved):
tested. looking good.

Connectors/Rhino/Speckle.Connectors.GrasshopperShared/Components/Operations/Send/SendAsyncComponent.cs
```diff
@@ -330,12 +330,14 @@ private void ParseInput(IGH_DataAccess da, IServiceScope scope)
...
         {
-          SpeckleGeometryWrapper wrapper = objWrapper.DeepCopy();
-          wrapper.Path = rootBase.Path;
-          wrapper.Parent = rootBase;
-          rootBase.Elements.Add(wrapper);
+          AddRuntimeMessage(
+            GH_RuntimeMessageLevel.Error,
+            "Speckle Geometry cannot be added directly to a Collection. "
+              + "Use a 'Speckle Data Object' component to wrap your geometry first, then pipe it into the Collection."
+          );
```
dogukankaratas, line 347: This early return exits before loop is closed, so the property keeps the stale collection from the previous send. A version keeps getting created even though it's an invalid one.

bjoernsteinhagen: fixed, good catch

dogukankaratas: looks good.

dogukankaratas (approved):

PR #1407, opened 2026-05-19
feat(grasshopper): expose root proxy objects as Proxies output on Load component

bjoernsteinhagen:
## Description

Users loading models from Revit, ETABS, and other host apps have been asking for access to the extra metadata those apps publish on the root object — things like `levelProxies`, `analysisResults`, etc. Up until now we were only outputting the collection and swallowing the extra stuff.

## User Value

Users can now access host-app-specific proxy data (levels, analysis results, etc.) directly from the Load component output.

## Changes:

- Added a static `Proxies` output to both the async and sync Load components (alongside the existing `Collection` and `Properties` outputs)
- `Proxies` is a `SpecklePropertyGroupGoo` keyed by proxy type name (e.g. `levelProxies`, `analysisResults`) — each value is a list of property groups, one per proxy instance. Wire it into Deconstruct to get discrete outputs per type, then Deconstruct again on individual items to get their fields
- Added `BuildProxiesGoo(Base root)` as a static helper on `SpeckleCollectionWrapper` — scans all dynamic `List<Base>` members on the root object, so it catches both known proxy types and anything custom a host app decides to publish, without us having to cherry-pick
- **No relational work is done on the proxies — they're surfaced as-is. First pass, users use Deconstruct to dig in**

Considered a dedicated `ExpandProxies` component (similar to `ExpandCollection`) but the simpler approach of a static output on Load more discoverable and avoids users needing to know to reach for an extra component.

## Screenshots & validation of changes:

image

## Checklist:


dogukankaratas:
Looks good to me. Only one comment: it's not possible to load analysis results in ETABS models. If this is a part of the request, it should be also added.

dogukankaratas (approved):

PR #1421, opened 2026-06-09
feat (tsd): member filtering

dogukankaratas:
Moves TSD send filtering from client-side to server-side.

Before, send fetched all members from TSD over gRPC (`GetMembersAsync(null)`) and filtered them client-side by GUID. 

Now the selection is keyed by `Type:Index`, and send asks TSD only for the selected member indices via `GetMembersAsync(indices)`.

### DIFF
--- a/Connectors/TSD/Speckle.Connectors.TSDShared/Utils/TsdObjectIdentifier.cs
+++ b/Connectors/TSD/Speckle.Connectors.TSDShared/Utils/TsdObjectIdentifier.cs
@@ -1,0 +1,24 @@
+using TSD.API.Remoting.Common;
+
+namespace Speckle.Connectors.TSDShared.Utils;
+
+internal static class TsdObjectIdentifier
+{
+  public static string Encode(EntityType entityType, int index) => $"{(int)entityType}:{index}";
+
+  public static (EntityType type, int index) Decode(string encodedId)
+  {
+    if (string.IsNullOrEmpty(encodedId))
+    {
+      throw new ArgumentException($"Invalid encoded ID: {encodedId}");
+    }
+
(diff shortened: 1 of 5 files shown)


Connectors/TSD/Speckle.Connectors.TSDShared/Utils/TsdObjectIdentifier.cs
```diff
@@ -0,0 +1,24 @@
...
+    if (string.IsNullOrEmpty(encodedId))
+    {
+      throw new ArgumentException($"Invalid encoded ID: {encodedId}");
+    }
+
+    var parts = encodedId.Split(':');
+    if (parts.Length != 2 || !int.TryParse(parts[0], out int typeValue) || !int.TryParse(parts[1], out int index))
+    {
+      throw new ArgumentException($"Invalid encoded ID: {encodedId}");
+    }
```
bjoernsteinhagen, line 20: if an ID in selection is malformed, entire send fails, no? Could this skip bad IDs with a warning log instead of throwing? or is this not even possible to get a "bad" or "stale" id?

bjoernsteinhagen (changes requested):

bjoernsteinhagen (approved):

Connectors/TSD/Speckle.Connectors.TSDShared/Utils/TsdObjectIdentifier.cs
dogukankaratas: can't really get a bad id here. we make these ids ourselves, not the user. a deleted member still parses fine and just gets skipped, no crash.

PR #1420, opened 2026-06-08
ci(github actions): lint & comply with linter findings

iainsproat:
## Description

Introduces a linter for GitHub Actions (zizmor), and updates the dependabot & GitHub Actions files to comply with Zizmor's recommendations.

## User Value

- CI only; improves security

## Changes:

- introduces a new job in the PR pipeline to lint github actions
- removes caching in release pipeline to prevent cache poisoning attacks
- pins all github actions to a commit sha to prevent supply-chain attacks
- introduces a 7 day cooldown period for dependabot updates to mitigate supply-chain attacks
- ignores the requirement to ensure all secrets are from environments. Ensures we can continue to use repository level secret
- temporarily disables codecov
  - CodeCov is currently breaking on GPG verification in our CI. This seems to be because CodeCov lost access to their GPG key account. https://github.com/codecov/codecov-action/issues/1956 Codecov have a patch, but that is less than our cooldown period. Out of caution, we'll disable CodeCov until the fix is older than our cooldown period.

## Validation of changes:

It works, see: https://github.com/specklesystems/speckle-sharp-connectors/actions/runs/27148670998/job/80133438281

Screenshot 2026-06-08 at 16 41 39

## Checklist:

- [ ] My commits are related to the pull request and do not amend unrelated code or documentation.
- [ ] I have added appropriate tests.
- [ ] I have updated or added relevant documentation.

PR #1434, opened 2026-06-17
fix(ci): fix release.yml semv ver env var

JR-Morgan:
The build step would inexplicitly output the semver as an the env var.
Because it wasn't explicit, it was broken with #1420

This PR fixes the logic, and makes the output explicit.

Also added a workflow to ensure that the test-installer CI pipeline is tested when changes are made to the workflows or build project

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,7 +10,7 @@ on:
 permissions: {} # We will use fine-grained permissions for each job, so we start with no permissions at all and then add them back as needed.
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow_ref }}-${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true # other running workflows get cancelled on the same branch
 
 jobs:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -58,11 +58,11 @@ jobs:
       - id: set-version
         name: Set version to output 
         run: |
-          echo "semver=${{ env.SEMVER }}"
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -50,15 +54,6 @@ jobs:
...
           compression-level: 0 # no compression
 
-      - id: set-version
-        name: Set version to output
-        run: |
-          echo "semver=${SEMVER}" >> "$Env:GITHUB_OUTPUT"
-          echo "file_version=${FILE_VERSION}" >> "$Env:GITHUB_OUTPUT"
-        env:
-          SEMVER: ${{ steps.get-version.outputs.semver }}
-          FILE_VERSION: ${{ steps.get-version.outputs.file_version }}
```
JR-Morgan, line 60: This logic was bad... There is no step called `get-version`, instead the build step would set it.

.github/workflows/release.yml
```diff
@@ -39,7 +39,11 @@ jobs:
       # we are purposefully not caching releases when publishing artefacts, see https://docs.zizmor.sh/audits/#cache-poisoning
 
       - name: ⚒️ Run build and zip connectors
-        run: ./build.ps1 zip
+        id: build-and-zip
+        run: |
+          ./build.ps1 zip
+          echo "semver=${SEMVER}" >> "$Env:GITHUB_OUTPUT"
+          echo "file_version=${FILE_VERSION}" >> "$Env:GITHUB_OUTPUT"
```
JR-Morgan: made it explicit

bjoernsteinhagen (approved):

PR #1437, opened 2026-06-22
fix(rhino): add volume and area to RhinoDataObject

dogukankaratas:
The BrepX → RhinoDataObject change dropped the area and volume values from properties. This adds them back into the object's properties as { name, value, units }.

image

### DIFF
--- a/Connectors/Rhino/Speckle.Connectors.RhinoShared/HostApp/Properties/PropertiesExtractor.cs
+++ b/Connectors/Rhino/Speckle.Connectors.RhinoShared/HostApp/Properties/PropertiesExtractor.cs
@@ -2,7 +2,9 @@ using Rhino;
 using Rhino.Collections;
 using Rhino.DocObjects;
 using Speckle.Connectors.Rhino.Mapper.Revit;
+using Speckle.Objects;
 using Speckle.Sdk;
+using Speckle.Sdk.Models;
 
 namespace Speckle.Connectors.Rhino.HostApp.Properties;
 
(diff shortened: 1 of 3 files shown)


JR-Morgan (approved):

Connectors/Rhino/Speckle.Connectors.RhinoShared/HostApp/Properties/PropertiesExtractor.cs
```diff
@@ -31,6 +33,29 @@ public PropertiesExtractor(RevitMappingResolver revitMappingResolver)
     return properties;
   }
 
+  public void AddGeometryProperties(Dictionary<string, object?> properties, Base geometry, string units)
+  {
+    if (geometry is IHasArea hasArea)
+    {
+      properties["area"] = new Dictionary<string, object?>
```
jsdbroughton, line 40: What if tyhe user already has a custom property called "area" ?

Connectors/Rhino/Speckle.Connectors.RhinoShared/HostApp/Properties/PropertiesExtractor.cs
```diff
@@ -31,6 +33,29 @@ public PropertiesExtractor(RevitMappingResolver revitMappingResolver)
...
+      {
+        ["name"] = "area",
+        ["value"] = hasArea.area,
+        ["units"] = $"{units}²",
+      };
+    }
+
+    if (geometry is IHasVolume hasVolume)
+    {
+      properties["volume"] = new Dictionary<string, object?>
```
jsdbroughton, line 50: What if the user already has a custome property called "volume"

PR #1442, opened 2026-06-26
fix(acad, c3d): composite hatch support

dogukankaratas:
Composite hatches (one hatch filling several separate areas) failed to send with "Composite hatches are not supported". The converter treated every loop as a hole. Now a separate area is united into the region instead of throwing.

Normal hatches are unaffected.


bimgeek (approved):
YOLO

PR #1443, opened 2026-06-29
fix(revit): Workaround for DLL conflict in Revit 2023 & 2024 with BimCAT plugin

JR-Morgan:
Bump `Microsoft.Data.Sqlite` from Version 7 to version 8 for Revit 2023 and Revit 2024 to have better compatibility with BimCAT plugin.

Jedd's Tested:<br>\[x\] 2025 (no changes required for us to be compatible)<br>\[x\] 2024 (bump sqlite)<br>\[x\] 2023 (bump sqlite)

Bilal's Tested:<br>\[x\] 2024<br>\[x\] 2023

---

Version 7 of Microsoft.Data.Sqlite is quite old, and isn't receiving security updates from Dotnet/efcore team.<br>I've wanted for a while to bump everything from 7 -> 8.

The only reason we pinned to 7.0.6 was that's what v2 used, and we used to care about side-by-side.<br>Now we care so little I think, that I'm not even going to test v2/v3 side by side to see if it's broken here. It may or may not.<br>Perhaps I'll test when we make the change for all connectors.

### DIFF
--- a/Connectors/Revit/Speckle.Connectors.Revit2023/Speckle.Connectors.Revit2023.csproj
+++ b/Connectors/Revit/Speckle.Connectors.Revit2023/Speckle.Connectors.Revit2023.csproj
@@ -34,6 +34,11 @@
   </ItemGroup>
 
   <ItemGroup>
+    <!--Better compatiblity with BimCAT plugin-->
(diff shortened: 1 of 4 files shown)


bimgeek (approved):
tested and they're all working.

PR #1445, opened 2026-07-01
chore(grasshopper): Migrate from mixpanel to posthog

JR-Morgan:
We're migrating our analytics from mixpanel to posthog.


Sdk/Speckle.Connectors.Common/Analytics/PostHogManager.cs
```diff
@@ -0,0 +1,162 @@
...
+      return;
+    }
+
+    if (string.IsNullOrWhiteSpace(account.userInfo.email))
+    {
+      throw new ArgumentException("Email cannot be empty.", nameof(account));
+    }
+
+    string[] semverBrake = application.SpeckleVersion.Split('-');
+    Version version = Version.Parse(semverBrake.First());
```
bjoernsteinhagen: can this throw?

Sdk/Speckle.Connectors.Common/Analytics/PostHogManager.cs
```diff
@@ -0,0 +1,162 @@
...
+    [CallerMemberName] string callerName = ""
+  )
+  {
+    if (!IsReleaseMode)
+    {
+      //only track in prod
+      return;
+    }
+
+    // Right now, we're keeping posthog only for app.speckle.systems users
```
bjoernsteinhagen, line 74: oki

bjoernsteinhagen (dismissed):

Sdk/Speckle.Connectors.Common/Analytics/PostHogManager.cs
JR-Morgan: If the version is not a semver, yes.
The `application.SpeckleVersion` is set from `AssemblyInformationalVersionAttribute`, which we are always setting as a valid semver.
I think it's desirable for this to fail loudly if that assumption ever gets broken in future, because it would fuck up metrics.

JR-Morgan: Maybe I move this to the constructor to make it harder to miss if ever we did break this rule

bjoernsteinhagen (approved):

PR #1454, opened 2026-07-08
fix(rhino): keep hatches as hatches on artefact round-trip

dogukankaratas:
Rhino hatches were sent as display meshes only and came back as plain meshes. Now a hatch is also kept as a raw 3dm blob.


bimgeek:
This has been addressed in https://github.com/specklesystems/speckle-sharp-connectors/pull/1455. Closing in favor of that.

PR #1456, opened 2026-07-08
fix: solids inside blocks receiving as meshes in Rhino/Grasshopper

oguzhankoral:
also need to go in: https://github.com/specklesystems/speckle-sharp-sdk/pull/502

A Brep/solid (or hatch) inside a block definition sent only its display mesh: the send stripped the 3dm SOLID blob from definition members, so DEFINES resolved to the mesh and the block rebuilt the solid as a mesh.

- Send: a definition member's 3dm solid is now emitted into DEFINES (no standalone SOLID edge — a member still renders only via a placed instance), and DEFINES ord counts per member so a member's solid + display mesh(es) stay grouped.
- Receive (Rhino + Grasshopper, both decode Rhino 3dm): group DEFINES by member ord and prefer the authoritative 3dm solid over its shadow mesh. A member with no solid keeps all its geometry; nested blocks recurse unchanged.

Viewer / AutoCAD / Revit are unaffected — they can't decode Rhino 3dm, so they fall back to the mesh (no duplicate). Also fixes hatches-inside-blocks. Depends on the SDK DefinesOrdByDefinition change (oguzhan/eng-8815-block-member-solids).

PS: the packages.lock.json churn is the local-SDK build pinning Microsoft.Bcl.AsyncInterfaces 9.0.10 (vs 9.0.4 in the published Speckle.Sdk.Dependencies). I think we will sort this later

PS2: fixes also unit isssue on receive

Screenshot 2026-07-09 at 00 55 59


bimgeek (approved):
tested. looking good.

PR #1458, opened 2026-07-09
feat(rhino): send + receive layer and object display colours

oguzhankoral:
Rhino was dropping both layer colours and by-object display colours on the 4.0 artefact round-trip.

Send (RhinoArtifactRootObjectBuilder):
- Layer CONTAINER nodes now carry the Rhino layer colour (argb) via the new AddCollection overload.
- Object display colours: RhinoColorUnpacker unpacks by-object colours (layers handled separately), emitted as COLOR nodes + HAS_COLOR edges, the same shape as materials.

Receive (RhinoHostObjectArtefactBuilder):
- Layers created from the scene view apply the resolved node colour.
- HAS_COLOR resolves to the owning object's appId and bakes as ObjectColor + ColorSource.ColorFromObject.

Both builders are the top-level artefact orchestrators; the colour additions pushed them over the CA1506 coupling ceiling, suppressed at the class level with justification (consistent with the existing method-level suppressions).

Lock files updated by restore against the local SDK.


bimgeek (approved):
tested. looking good.

PR #1462, opened 2026-07-20
fix(civil3d): send corridor geometry in the 4.0 artefact path

dogukankaratas:
Corridors sent on the 4.0 path arrived properties-only (no geometry, invisible, false SUCCESS): the converter hung solids/feature lines on plain `Base` children, but the builder only walks the typed `Civil3dObject` tree.

Fix: corridor children are now typed `Civil3dObject`s throughout (baseline → featurelines + regions → assembly → subassemblies), so solids and feature-line polylines emit as DISPLAY and the SUBELEMENT tree stays intact. Builder now labels subelement nodes by real name/type. Civil3D 2023 unaffected (existing `CIVIL3D2024_OR_GREATER` guard).


bimgeek (approved):
tested. can confirm its fixed.

PR #1464, opened 2026-07-20
feat(groups): via IN_GROUP (17) for Rhino + AutoCAD

oguzhankoral:
Send: both artifact builders collect group membership Base-free (no GroupProxy) — Rhino from each object's GetGroupList (overlap = implicit nesting), AutoCAD from persistent reactors — and emit one CONTAINER(subtype "Group") node + IN_GROUP membership edges per group. A separate axis from IN_COLLECTION: an object keeps its layer AND its group(s). Definition members stay suppressed (no scene edges), and a group emptied by that is skipped.

Receive: a new Groups phase inverts GroupsByObject to group -> baked member ids (block placements included) and rebuilds native groups — Rhino group table / AutoCAD group dictionary (own committed doc-TM transaction; unique keys get a -N suffix). Names carry the baseLayerName suffix like the v1 bakers, and pre-clean now purges them, so re-receive is idempotent (fixes the old AutocadGroupBaker TODO).

Pairs with SDK big-truck (pipeline.InGroup / ArtefactRelations. GroupsByObject) + bundle-spec main (rel 17 un-retired, live). Tested live in Rhino and AutoCAD both directions.


bimgeek:
tested for Rhino. it works as expected.

bimgeek:
autocad is good too.

bimgeek (approved):

PR #1467, opened 2026-07-21
fix(civil3d): round-trip property sets on 4.0 artefact versions

dogukankaratas:
Property sets stopped round-tripping on Civil3D 4.0 (artefact) versions. Two breaks, both fixed:

- **Send:** the shared artefact builder never emitted the property-set *definitions* (only per-object values reached EAV). Added a `Civil3dArtifactRootObjectBuilder` that writes the definitions the converter accumulates.
- **Receive:** 4.0 versions bake through the shared AutoCAD builder, so no Civil-specific baking ran. Added a `Civil3dHostObjectArtefactBuilder` that purges → recreates definitions → attaches sets.


oguzhankoral:
fixed couple of things and merging now. tested with civil3d 2026

oguzhankoral (approved):

PR #1474, opened 2026-07-22
Round-trip the Revit reference point through the artefact bundle (ENG-8808)

kekesidavid:
## What

Fixes **ENG-8808**. The Speckle 4.0 artefact send path dropped the reference-point transform the v1 pipeline stored on the root object, so Revit→Revit reference-point round-tripping (send Project Base / Survey → receive Source) was lost for every model sent through the artefact path — geometry landed offset by the base-point vector.

This records the applied reference point in the bundle `meta` (**ENG-8947**) and reverses it on receive.

## Changes

**Send** — `RevitArtifactRootObjectBuilder`
- Writes `reference_point_kind` + `reference_point_offset` (display units, the vector subtracted) for the **main (non-linked)** model: `projectBasePoint` / `surveyPoint`, or `internalOriginFallback` when the requested base point is missing.
- Only the host document's transform is persisted — a linked doc's is `(referencePoint ∘ linkPlacement⁻¹)` and must not be.
- The requested `ReferencePointType` is threaded to the send pipeline via `RevitConversionSettings` (the resolved transform alone can't distinguish `projectBasePoint` from `surveyPoint`).

**Receive**
- **net8** (`RevitHostObjectArtefactBuilder`): rebuilds the translation from the meta offset (scaled to internal feet), composes it with the local receive setting via the shared `ReferencePointHelper.CalculateNewTransform`, and applies through `ConvertToInternalCoordinates`.
- **net48**: works through the SDK reconstruction lift — `RevitHostObjectBuilder` is unchanged (its private `CalculateNewTransform` just moved to the shared helper).

## Shared Coordinates — deliberately NOT round-tripped

The Revit connector has a 4th reference-point kind, **Shared Coordinates** (`ActiveProjectLocation.GetTotalTransform()`), which:
- is **outside** ENG-8947's vocabulary (`internalOrigin | projectBasePoint | surveyPoint | internalOriginFallback`), and
- can carry **true-north rotation**, which a translation-only `reference_point_offset` structurally cannot represent, and
- has **no counterpart in the converters/native pipeline** (`rvextract` only reads the survey-point *position*; it would reject a `sharedCoordinates` value).

So Shared Coordinates records **no meta** and does **not** round-trip — a deliberate scope call (Option B): rather than invent a `sharedCoordinates` value nobody in the shared bundle-spec consumes, or mislabel it as `surveyPoint`, we leave it out. Internal Origin / Project Base / Survey are pure translations and map to ENG-8947 exactly. If Shared-coords georeferencing/federation later becomes a requirement, `sharedCoordinates` (with rotation handling) should be added to the bundle-spec **and** both pipelines together.

## Requires

The `meta` provision in the SDK: specklesystems/speckle-sharp-sdk#511

## Test

Revit 2025/2026: model with a Project Base Point offset from the internal origin → send with **Reference Point = Project Base** (also **Survey**) → receive into a second model (**Reference Point = Source**) → geometry should land on the source's internal coordinates (matching `dev`). Verify **Internal Origin** is unchanged and **linked models** still place correctly.

### DIFF
--- a/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
+++ b/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
@@ -183,7 +183,8 @@ internal sealed class RevitSendBinding : RevitBaseBinding, ISendBinding
               _toSpeckleSettingsManager.GetLinkedModelsSetting(document, card),
               _toSpeckleSettingsManager.GetSendRebarsAsVolumetric(document, card),
               _toSpeckleSettingsManager.GetSendAreasAsMesh(document, card),
-              false
+              false,
+              referencePointKind: _toSpeckleSettingsManager.GetReferencePointKind(card)
             )
           );
       },
--- a/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Receive/RevitHostObjectArtefactBuilder.cs
+++ b/Connectors/Revit/Speckle.Connectors.RevitShared/Operations/Receive/RevitHostObjectArtefactBuilder.cs
@@ -10,6 +10,7 @@ using Speckle.Connectors.Common.Threading;
 using Speckle.Connectors.Revit.HostApp;
 using Speckle.Converters.Common;
 using Speckle.Converters.RevitShared;
+using Speckle.Converters.RevitShared.Helpers;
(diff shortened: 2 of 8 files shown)


oguzhankoral (approved):
tested and looks good to me. i suggest to double test when @bimgeek is back

PR #1481, opened 2026-07-29
feat(rhino): Rhino 9/net10 head, headless connector slice, Build/BuildAndUpload split (ENG-8980)

gjedlicska:
Linear: [ENG-8980](https://linear.app/speckle/issue/ENG-8980/big-truck-upstream-rhino-9net10-head-headless-connector-slice-build) — leg 1 of the Rhino leg productionization spec (`atlas/specs/2026-07-rhino-leg-productionization.md`).

The three upstream pieces the containerized Rhino converter leg consumes, landed as the compile-proven shapes from the wayfinder research:

## `Converters/Rhino/Speckle.Converters.Rhino9`
net10 head importing the shared Rhino converter projitems **unmodified** (the shared source has zero `#if RHINO*`). Exact pins in the head csproj — no floating ranges; they move in lockstep with the converters image's rhino3d deb pin:
- `RhinoCommon 9.0.26167.11545-wip`
- `Rhino.Inside 9.0.26084.13070-beta` (flows transitively so the host exe inherits the pin)
- `Grasshopper 9.0.26167.11545-wip` — not used by the converter; required because Rhino.Inside's open-range Grasshopper floor exact-pins an older RhinoCommon, which trips NU1608-as-error

## `Connectors/Rhino/Speckle.Connectors.RhinoHeadless`
The DUI3-free slice around `RhinoArtifactRootObjectBuilder` — 15 files linked from `RhinoShared` (builder + unpackers + `PropertiesExtractor` + Mapper/Revit + Extensions) with the `Speckle.Connectors.Common` closure. Two csproj shims, no source changes:
- `<Using Include="System.Drawing" />` (Windows heads get it via `UseWindowsForms`; only `Color` is used — cross-platform)
- `<Using Remove="System.IO" />` (the WPF heads drop `System.IO` from implicit usings, so the shared files carry explicit `using System.IO;` — a plain net10 project must mirror that view or IDE0005-as-error flags them)

## `Build` / `BuildAndUpload` split
`RhinoArtifactRootObjectBuilder` gains a public `Build(objects, projectId?, versionId, outputDir, progress, ct)` → new `ArtifactBundleResult` (bundle files, synthetic root id, object count, conversion results): collect + write-bundle-to-outdir, touching **no auth, storage, or server API**. `BuildAndUpload` calls the same core then uploads — `IArtifactRootObjectBuilder<T>` is unchanged and the other six implementers are untouched, so connector consumers behave exactly as before. This lets the converter host stop at the outdir; dispatch stays the single uploader.

Both new projects are added to `Speckle.Connectors.slnx` / `Local.slnx` / `Speckle.Rhino.slnx` under `/Connectors/Rhino/9/`.

## Verification
- Head + slice build on linux-x64 / net10, `-c Local` against the sibling sdk checkout: **0 warnings, 0 errors**
- Existing consumers of the shared builder still compile: `Speckle.Connectors.RhinoImporter` (net8.0-windows) and `Speckle.Connectors.Rhino8` (net48)
- All 137 tests pass (`Connectors.Common.Tests`, `Converters.Common.Tests`, `DUI.Tests`, `-c Local`)
- Note for reviewers: `-c Local` currently needs the sibling sdk at `ceb47b7` — sdk `4e043bd` (material PBR channels #516) changed `ObjectsArtifactPipeline.AddMaterial` and breaks the connectors build until connectors adapt. Release CI is red on big-truck independently of this PR (published `Speckle.Objects` lacks the big-truck pipeline APIs).

### DIFF
--- a/Converters/Rhino/Speckle.Converters.Rhino9/Speckle.Converters.Rhino9.csproj
+++ b/Converters/Rhino/Speckle.Converters.Rhino9/Speckle.Converters.Rhino9.csproj
@@ -1,9 +1,9 @@
 <Project Sdk="Microsoft.NET.Sdk">
-  <!-- Rhino 9 WIP head for the headless linux converter leg (rhino.inside, net10). Consumed by the
-       speckle-converters host via sibling-checkout project references (Local configuration).
+  <!-- Rhino 9 WIP head for the headless linux converter leg (rhino.inside, net10), consumed by the
+       speckle-converters host via sibling-checkout project references.
        EXACT version pins only: the RhinoCommon/Rhino.Inside pins move in lockstep with the converters
-       image's rhino3d deb pin (monthly WIP rebuild) — a float on either side silently skews the managed
-       API surface against the installed Rhino and fails at runtime. -->
+       image's rhino3d deb pin — a float on either side silently skews the managed API surface against
+       the installed Rhino and fails at runtime. -->
   <PropertyGroup>
     <TargetFramework>net10.0</TargetFramework>
     <Configurations>Debug;Release;Local</Configurations>
--- a/Converters/Rhino/Speckle.Converters.Rhino9/Speckle.Converters.Rhino9.csproj
+++ b/Converters/Rhino/Speckle.Converters.Rhino9/Speckle.Converters.Rhino9.csproj
@@ -1,0 +1,37 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <!-- Rhino 9 WIP head for the headless linux converter leg (rhino.inside, net10). Consumed by the
+       speckle-converters host via sibling-checkout project references (Local configuration).
+       EXACT version pins only: the RhinoCommon/Rhino.Inside pins move in lockstep with the converters
+       image's rhino3d deb pin (monthly WIP rebuild) — a float on either side silently skews the managed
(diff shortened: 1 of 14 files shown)


Converters/Rhino/Speckle.Converters.Rhino9/Speckle.Converters.Rhino9.csproj
```diff
@@ -0,0 +1,37 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <!-- Rhino 9 WIP head for the headless linux converter leg (rhino.inside, net10). Consumed by the
+       speckle-converters host via sibling-checkout project references (Local configuration).
+       EXACT version pins only: the RhinoCommon/Rhino.Inside pins move in lockstep with the converters
+       image's rhino3d deb pin (monthly WIP rebuild) — a float on either side silently skews the managed
+       API surface against the installed Rhino and fails at runtime. -->
```
JR-Morgan: Not a fan of these code comments.
It's talking about things that are delocalised from this file, It's very easy for these to become out of date and I don't think they're very important.

Claude has some bad habits r.e. code comments that I've been noticing in other repos. Might be worth telling it to remove any unimportant ones and simplify the reset to single line where possible.

JR-Morgan (approved):

Issue #503, opened 2025-01-18
🚨 Beam Connections Lost: Revit→Speckle→ETABS Sync Corrupts Z-Elevation & Joint Behavior


archeduardo5:
## Prerequisites

- [x] I read the [contribution guidelines](https://github.com/specklesystems/speckle-server/blob/main/CONTRIBUTING.md)
- [x] I checked the [documentation](https://speckle.guide/) and found no answer.
- [x] I checked [existing issues](../issues?q=is%3Aissue) and found no similar issue. 
- [x] I checked the [community forum](https://speckle.community/) for related discussions and found no answer.
- [x] I'm reporting the issue to the correct repository (see also [speckle-server](https://github.com/specklesystems/speckle-server), [speckle-sharp](https://github.com/specklesystems/speckle-sharp), [specklepy](https://github.com/specklesystems/specklepy), [speckle-docs](https://github.com/specklesystems/speckle-docs), and [others](https://github.com/orgs/specklesystems/repositories))

## System Info

If applicable, please fill in the below details - they help a lot!

### Smartphone:
- Device: iPhone 11 Black
- OS: [por favor indica tu versión de iOS]
- App: GitHub App
- Version: 2025.4

## Failure Logs
```text

<!---
Please include any relevant log snippets or files here, or upload as a file.

If including inline, please use markdown code block syntax. https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks
For example:

```
your log output here
```
-->

## Additional context

<!---
Add any other context about the problem here.
-->

## Proposed Solution (if any)

<!---
Let us know what how you would solve this.
-->

#### Optional: Affected Projects

<!---
Does this issue propagate to other dependencies or dependents? If so, list them here with links!

bjoernsteinhagen:
Hey @archeduardo5,

Please provide some more information as to the reported bug. What Z-elevations are being "corrupted" and by joint behaviour are you referring to the end releases assignments being exchanged between Revit analytical and ETABS?

Some screenshots would also help greatly. 

(Note: this repo is for our Next-Gen connectors, I assume this issue pertains to the v2 connectors, in this repo: [speckle-sharp](https://github.com/specklesystems/speckle-sharp))

archeduardo5:
[...]

PR #1489, opened 2026-07-30
fix(autocad): text survives send + receive (ENG-8827)

kekesidavid:
Closes [ENG-8827](https://linear.app/speckle/issue/ENG-8827/autocad-send-text-dbtextmtext-publishes-geometry-less).

## The ticket's premise was stale

`SgeoEncoder` gained a **Text primitive** (code 12) in SDK specklesystems/speckle-sharp-connectors#503 on 9 July, and the viewer got its decode the same day — so DBText/MText *do* publish geometry now. Verified against a real bundle in `%TEMP%\Speckle\artifacts`: `geometries.parquet` carries `text` rows with correct values, planes and `DISPLAY` rels. Three separate defects sat downstream of that, which is why text still vanished end to end.

## 1\. Receive dropped every text object

```
geom 112 (SGEO) convert of Objects.Annotation.Text failed — ConversionException:
No direct conversion found for type Speckle.Objects.Annotation.Text
```

`ToHost/Geometry/TextToHostConverter.cs` existed on disk but was **never listed in** `Speckle.Converters.AutocadShared.projitems`. Shared projects don't glob, so it compiled into no assembly, registered nowhere, and failed only at runtime. Listing it is the entire receive fix.

## 2\. MText published `maxWidth = 0`

AutoCAD spells "no wrap" as `Width == 0`; Speckle spells it `null`. Civil3D corridor labels went out as `flags=1024, maxWidth=0.0`. The viewer's `TextBatch` does `maxWidth !== null ? maxWidth : Infinity`, handing troika a zero-width column — and because the batch resolves on a shared sync counter, one degenerate entry can take the whole text batch down with it, hiding the otherwise well-formed DBText too.

Now publishes `null`, and receive treats a non-positive `maxWidth` as unset so **already-published** versions still bake correctly.

## 3\. Justified DBText anchored on the wrong point

It used `Position` — stale for anything but left-justified text — with hard-coded `Left`/`Bottom`. Now anchors on `AlignmentPoint` when justified, and maps `Justify` through a new `TextAlignmentMap` shared with the MText converter (which previously carried a private copy of the same mapping).

## The ticket's EAV half

New `TextPropertiesExtractor` puts value / height / position / rotation / style under a `Text` group — plus justification, width factor and oblique for DBText; attachment, wrap width and the raw markup for MText; `tag` for block attributes. Wired into the AutoCAD, **Civil3D and Plant3D** extractors, since a Civil drawing's annotation goes through Civil3D's own `PropertiesExtractor`. Text content is now queryable, not only encoded in the geometry blob.

Also on receive: plain text is escaped into MText contents (`\`, `{`, `}` are MText control characters) and newlines become `\P`, so multi-line text bakes as paragraphs instead of literal markup.

## Testing

* `Speckle.Connectors.Autocad2025`, `Speckle.Connectors.Civil3d2026` and `Speckle.Connectors.Plant3d2026` build `-c Local` with 0 warnings / 0 errors; `dotnet csharpier check ./` clean; no lock-file churn.
* Verified `TextToHostConverter`, `TextPropertiesExtractor` and `TextAlignmentMap` are present in the deployed converter assemblies.
* Host-app verification (send → viewer, receive into a blank DWG) is being done locally against these builds.

## Follow-up in another repo

`TextBatch.ts:359` should treat `maxWidth <= 0` as no-wrap, not just `null`. Without that, every model already published with `maxWidth = 0` stays broken in the viewer even after this fix. Left out of scope here.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

### DIFF
--- a/Converters/Autocad/Speckle.Converters.AutocadShared/Helpers/TextAlignmentMap.cs
+++ b/Converters/Autocad/Speckle.Converters.AutocadShared/Helpers/TextAlignmentMap.cs
@@ -1,0 +1,74 @@
+namespace Speckle.Converters.Autocad.Helpers;
+
+/// <summary>
+/// Maps an AutoCAD <see cref="ADB.AttachmentPoint"/> onto the 3x3 alignment grid Speckle's
+/// <see cref="SA.Text"/> exposes. Shared by the MText converter (where it comes from
+/// <c>MText.Attachment</c>) and the DBText converter (where it comes from <c>DBText.Justify</c>),
+/// so both spell the same anchor semantics.
+/// </summary>
+public static class TextAlignmentMap
+{
+  /// <summary>
+  /// Simplify horizontal text alignment to 3 options: Left, Center, Right.
+  /// </summary>
+  public static SA.AlignmentHorizontal GetHorizontalAlignment(ADB.AttachmentPoint attachmentPt) =>
+    attachmentPt switch
+    {
+      ADB.AttachmentPoint.TopLeft
+      or ADB.AttachmentPoint.MiddleLeft
+      or ADB.AttachmentPoint.BottomLeft
+      or ADB.AttachmentPoint.BaseLeft => SA.AlignmentHorizontal.Left,
+      ADB.AttachmentPoint.TopCenter
+      or ADB.AttachmentPoint.MiddleCenter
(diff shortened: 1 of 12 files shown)


bimgeek (approved):
tested and approved.

PR #1492, opened 2026-07-30
fix(rhino): resolve instance definitions without the doc-level instance C-API (ENG-9047)

gjedlicska:
## Problem

On the headless Linux rhino leg (speckle-converters, rhino.inside against the Rhino 9 WIP deb), every block instance fails to unpack: `InstanceObject.InstanceDefinition` throws `EntryPointNotFoundException` because the Linux native (`librhcommon_c.so`) ships **none** of the doc-level instance C-API — `CRhinoInstanceObject_*`, `CRhinoInstanceDefinition_*`, `CRhinoInstanceDefinitionTable_*` are all absent (verified on debs 9.0.26167 and 9.0.26211; the bundled RhinoCommon.dll binds them anyway). An all-blocks model fails wholesale.

[ENG-9047](https://linear.app/speckle/issue/ENG-9047)

## Fix

`RhinoInstanceUnpacker` now:

- reads the instance transform + definition id off the `ON_InstanceRef` geometry (`InstanceReferenceGeometry.ParentIdefId`/`.Xform` — same underlying data, and the opennurbs C-API **is** implemented on Linux);
- keeps `instance.InstanceDefinition` as the primary path for definition name/description/members, and on the first `EntryPointNotFoundException` flips a process-wide flag and resolves definitions from the doc's source 3dm instead (`File3dm.Read` filtered to the instance-definition table, member ids mapped back through `doc.Objects.FindId`).

Windows/Mac natives implement the doc-level API, so the fallback is unreachable there; the doc-level path is byte-identical to before. All APIs used exist down to RhinoCommon 7.13 (the Rhino7 head's pin), so every head compiles.

## Known limits / notes

- Linked (non-embedded) definitions resolve empty under the fallback (the file table entry carries the link path, not member ids) — noted in-code; irrelevant to the converter leg's corpus, unreachable elsewhere.
- `RhinoMapperBinding` still calls `InstanceObject.InstanceDefinition` in 4 places — interactive DUI binding code, not on the headless path, left untouched.

## Verification

Empirically verified in the converter image (probe + full tripwire tier): `blocks-nested.3dm` (nested defs, 5 instances) converts 8/8 objects and schema-validates; `attributes-eav.3dm` objectsFailed dropped 1 → 0; all 7 positive fixtures pass.

speckle-converters pins this commit as its vendor gitlink on the ENG-9047 branch — please prefer merge over squash/rebase, or the gitlink needs a re-pin to the landed SHA.

### DIFF
--- a/Connectors/Rhino/Speckle.Connectors.RhinoShared/HostApp/RhinoInstanceUnpacker.cs
+++ b/Connectors/Rhino/Speckle.Connectors.RhinoShared/HostApp/RhinoInstanceUnpacker.cs
@@ -1,6 +1,7 @@
 ﻿using Microsoft.Extensions.Logging;
 using Rhino;
 using Rhino.DocObjects;
+using Rhino.FileIO;
 using Rhino.Geometry;
 using Speckle.Connectors.Common.Instances;
 using Speckle.Connectors.Rhino.Extensions;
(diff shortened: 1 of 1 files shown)


Connectors/Rhino/Speckle.Connectors.RhinoShared/HostApp/RhinoInstanceUnpacker.cs
```diff
@@ -133,6 +151,61 @@ out List<InstanceProxy>? instanceProxiesWithSameDefinition
...
+    {
+      throw new SpeckleException(
+        "Cannot resolve instance definitions: the doc-level instance API is unavailable on this platform and the doc has no source 3dm to read them from"
+      );
+    }
+
+    // Known limit (ENG-9047): a linked (non-embedded) definition's file table entry carries
+    // the link path, not member ids, so it resolves to an empty definition here — where the
+    // doc-level GetObjects() would return the loaded linked geometry.
+    using var file = File3dm.Read(doc.Path, File3dm.TableTypeFilter.InstanceDefinition, File3dm.ObjectTypeFilter.None);
```
oguzhankoral, line 189: If it works 🤷‍♂️
It is a legit temp fallback

JR-Morgan (approved):

PR #1496, opened 2026-08-04
fix(revit): receive Rhino solids as solids, not meshes (ENG-8800)

kekesidavid:
Closes [ENG-8800](https://linear.app/speckle/issue/ENG-8800/rhino-solids-received-in-revit-as-meshes-raw-3dm-solid-blob-dropped-on).

Receiving a Rhino model with solids into Revit baked tessellated mesh DirectShapes instead of the real solids `dev` produces: the lossless raw 3dm `SOLID` blob that the Rhino send side writes *alongside* the display meshes was dropped on the receive side. Send is fine (`RhinoArtifactRootObjectBuilder` emits both).

## Both Revit receive paths dropped it

Not for the reason the ticket assumed — the split isn't net48 vs net8 (`RevitHostObjectArtefactBuilder` is registered unconditionally for 2023–2027), it's which builder runs:

- **direct bake (default):** `DecodeGeometry` only read SGEO mesh blobs and never consulted `rels.SolidByObject`; `BakeAtomic` also skipped solid-only objects as non-geometric. Rhino/AutoCAD/Grasshopper all consult it — Revit was the outlier.
- **reconstruction (`ReceiveInstancesAsFamilies` on):** `ArtifactReceiveOptions(PreferSolids: false)` made `ObjectsArtifactReader` strip the `rawEncoding` before `IRawEncodedObjectConverter` could import it. The registration comment claiming "Revit can't import 3dm" was wrong — `dev` demonstrably imports it via `DB.ShapeImporter`.

## Approach

Prefer the solid, but don't commit to it — mirroring AutoCAD's ENG-8820 semantics. An import that returns nothing, degrades to `DB.Mesh`, or throws falls back to the display meshes, which exist precisely as that shadow. Non-`RHINO_3DM` blobs are filtered out up front so a foreign encoding (e.g. AutoCAD SAT) never reaches the importer.

- **new `RevitArtefactSolidImporter`** — `DB.ShapeImporter` in a `SubTransaction` with rollback-on-empty (a failed import leaves ghost state that breaks painting and grouping, same as the v1 converter), the composed reference-point transform applied via `SolidUtils.CreateTransformed` (mesh vertices get this through `ToInternalPoints`; raw imports otherwise wouldn't — ENG-8808/ENG-8947), and post-commit face painting since imported solids carry no material of their own.
- **`RevitHostObjectArtefactBuilder`** — solid preferred per atomic object and per definition member (grouped by `DefinesOrdByDefinition`, as in `RhinoHostObjectArtefactBuilder`), gate widened to `rels.SolidByObject`, paint targets collected during the bake then painted in their own transaction once it commits (an element's faces aren't queryable before then — same ordering as v1's `PostBakePaint`).
- **`RevitConnectorModule`** — `PreferSolids: true`, fixing the families-receive path, plus the corrected comment.

## Verification

Real Rhino→Revit 2026 receive on this build, from the artefact session log:

```
model: rhino/solids
--- counters ---   solidsImported: 23   solidImportDegradedToMesh: 3   nonGeometricSkipped: 4
```

23 blobs became real Revit solids (individually paintable faces, confirmed with the Paint tool); 3 degraded imports fell back to display meshes by design. Full `Local.slnx -c Local` build after a `deep-clean-local`: 0 errors, 0 warnings; csharpier clean.

New session-log counters report which path each object took: `solidsImported`, `solidImportEmpty`, `solidImportDegradedToMesh`, `solidImportNotTransformable`, `solidImportFailed`, `solidsPainted`, `solidsUnpainted`, `solidPaintFailed`.

## Not in this PR

- **Block/instance placements.** A solid inside a `DB.GeometryInstance` can't have its faces painted or individually selected, so block geometry still isn't face-paintable. Flattening for both receive paths exists on `david/eng-8800-instance-solid-flattening` but is unverified: the nested-blocks version I tested carries **no 3dm blobs at all** (`geometryBlobs == definitions`, zero solid counters), so whether the Rhino artefact send emits member solids for block definitions is still open.
- **Material on imported solids.** The verification run queued no paint targets (`solidsPainted`/`solidsUnpainted` absent), so no material resolved for those 23 solids — worth a look if the source objects do carry materials.
- **Polylines** error out on the artefact bake path ("did not convert to any native geometry") — `DecodeGeometry` handles SGEO meshes only. Pre-existing gap, separate ticket.

### DIFF
--- a/Connectors/Revit/Speckle.Connectors.RevitShared/DependencyInjection/RevitConnectorModule.cs
+++ b/Connectors/Revit/Speckle.Connectors.RevitShared/DependencyInjection/RevitConnectorModule.cs
@@ -89,16 +89,19 @@ public static class ServiceRegistration
     serviceCollection.AddScoped<ITransactionManager, TransactionManager>();
 
     // Speckle 4.0 artefact receive: download the parquet bundle + reconstruct the Base graph (DataObjects with
-    // displayValue meshes → DirectShapes). Revit can't import 3dm, so PreferSolids = false (meshes only).
+    // displayValue meshes → DirectShapes). PreferSolids = true: reconstruction only runs when receive-as-families is on
+    // (RevitHostObjectArtefactBuilder.Build delegates to the v1 builder), and Revit DOES import 3dm — the rebuilt
+    // RhinoObject.rawEncoding goes through IRawEncodedObjectConverter → DB.ShapeImporter → real solids [ENG-8800].
     serviceCollection.AddScoped<
       Speckle.Sdk.Pipelines.Receive.Artifacts.IArtifactDownloader,
       Speckle.Sdk.Pipelines.Receive.Artifacts.ArtifactDownloader
     >();
-    serviceCollection.AddSingleton(new Speckle.Objects.Utils.ArtifactReceiveOptions(PreferSolids: false));
+    serviceCollection.AddSingleton(new Speckle.Objects.Utils.ArtifactReceiveOptions(PreferSolids: true));
     serviceCollection.AddScoped<IArtifactReceiver, ArtifactReceiver>();
-    // Native artefact receive (DirectShape from SGEO meshes, Base-free). Registering this activates the direct-bake
+    // Native artefact receive (DirectShape from raw 3dm solids / SGEO meshes, Base-free). Registering this activates the direct-bake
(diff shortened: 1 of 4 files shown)


bimgeek (approved):
tested. can confirm the fix.

PR #1498, opened 2026-08-05
fix(rhino): close six artefact round-trip gaps (ENG-9108, 9111, 9112, 9113, 9114, 9115)

kekesidavid:
Six of the nine Rhino artefact tickets in the Speckle 7.0 backlog, one commit each. Three are skipped and explained at the bottom.

All of these are round-trip gaps in the 4.0 rewrite rather than new features — data that either never made it into the bundle or was in the bundle and never read back out.

## Fixed

| Ticket | What was wrong |
|---|---|
| **ENG-9108** (High) | An object with `MaterialFromLayer` published with no material at all. The unpacker's object stage skips those by design and its layer stage builds a proxy listing the **layer** id — but phase 2 resolved proxy members only through `geometryKsByObjectId`, which is keyed by object id, so the MATERIAL node landed with zero `HAS_MATERIAL` edges. Phase 1 now records object → inherited-layer id (only when that layer really carries a material, mirroring the condition the unpacker uses, so every recorded object resolves); phase 2 inverts that and expands a layer-sourced proxy onto its inheritors' display geometry. |
| **ENG-9111** (High) | `ArtefactBundle.Properties` was never applied to object attributes, so Rhino user text / user dictionaries didn't survive a round trip and Revit + IFC parameters never reached received objects. The data was in the bundle the whole time. New `RhinoArtefactUserStrings` flattens it to the legacy dot-delimited user-string shape, for atomic objects and instance placements. |
| **ENG-9112** | Named views were being written to `envelope.camera_views.parquet` and surfaced as `ArtefactBundle.CameraViews`, but the bake never read the list. New `RhinoArtefactViewBaker` as step 6. |
| **ENG-9113** | One source object decoding to several geometries baked as loose objects, so a Revit wall arrived as a pile of meshes. Now bound into a native group. |
| **ENG-9114** | A colour set directly on a block instance is emitted object-sourced (`ord=1`, ENG-8825/8822) and parsed into `ColorByObject`, but `CreateColors` read only `ColorByGeometry` — so placements were never in the map. |
| **ENG-9115** | `DeepClean` purged this model's groups and layers but not its instance definitions, so every receive left another orphaned `…-(def-K)` set in the block table. |

### Notes on the less obvious bits

**ENG-9112** deliberately keeps the v1 `RhinoViewBaker` mechanism (clone the active viewport, retarget, push, add, restore) — going through a real viewport is what supplies a valid screen port and frustum aspect, which a bare `new ViewInfo()` does not have. Two things it does differently:

- **Projection type is preserved.** v1 unconditionally called `ChangeToPerspectiveProjection`, so parallel views came back perspective. `IsOrtho` is recorded, so a parallel view now switches to parallel and rebuilds a symmetric frustum from the recorded `OrthoHeight` + aspect.
- **Lens goes through `Camera35mmLensLength`.** The `lensLength` argument of `ChangeToPerspectiveProjection` is documented to be ignored when the viewport is already perspective, which a cloned active viewport normally is — so v1's value never took effect there. (v1 also passed `1.0` into it, having the argument order the other way round in its comment.)

**ENG-9113** matches the v1 contract exactly: the report entry points at the **group** id, because `RhinoBasicConnectorBinding.GetObjectsFromIds` resolves an id through both `Objects.FindId` and `Groups.FindId`, so highlighting from the report selects the whole element. `bakedObjectIds` keeps only the member guids — a group id there would make whole-model highlighting walk every object of every group, which is why v1 kept them out too.

**ENG-9115** deletes with `deleteReferences: true`: the previous receive's placements are still live at that point (their layers are purged immediately after), and with `false` Rhino refuses to delete while any reference exists, leaving exactly the orphans this removes. Definition names are matched on the **cleaned** suffix, since `CleanBlockDefinitionName` rewrites `/` and `\` and a project or model name containing either would otherwise not match.

**ENG-9111** also lists `RhinoArtefactUserStrings.cs` in the `RhinoShared` `.projitems` — that shared project enumerates `<Compile>` entries explicitly, so an unlisted file compiles into nothing. Same trap as ENG-8827 on the Autocad side.

## Scope limits, called out

- **ENG-9111 does not cover block-definition members.** A member gets no `DISPLAY` edge by design (it renders only through its definition) and `ObjectByGeometry()` is built from `DISPLAY` alone, so on receive there's no route from a definition's geometry back to the member object holding the eav row. The ticket's acceptance criteria don't require it. If ENG-9110 introduces a member↔geometry association, member properties come along nearly for free.
- **ENG-9108** skips a block instance inheriting from its layer: it owns no geometry, and instance-sourced `HAS_MATERIAL` needs the namespace tag `HAS_COLOR` got in ENG-8822 — that's ENG-9109.

## Skipped, and why

- **ENG-9107** (text publish) — the ticket's root cause is stale. `Speckle.Objects`/`Speckle.Sdk` resolve as a **`ProjectReference`** to the sibling `speckle-sharp-sdk` checkout under the `Local` configuration, and that checkout (also on `big-truck`) already has `Text`/`Region` in both `SgeoEncoder` and `SgeoDecoder`. There's also no package to bump — the pinned 2026.6.0 is the newest on nuget.org, and it predates the artefact pipeline entirely. Rhino's four text converters all exist and register, and the Rhino converter `.projitems` uses a `**\*.cs` wildcard, so nothing is silently unlisted. Needs a publish from a real Rhino document to see whether text still errors.
- **ENG-9109** (instance materials) — the spec allows `geometry|instance`, but `ObjectsArtifactPipeline.HasMaterial(geometryK, materialK)` has no namespace-tag parameter and the reader files every `HAS_MATERIAL` into `MaterialByGeometry`. Needs SDK work plus a convention decision (tag the INSTANCE node K per the spec text, or the object K as the colour path does).
- **ENG-9110** (member layers) — `ON_LAYER` exists only as prose in `docs/`. It's not in `speckle-bundle-spec/spec/bundle-spec.sql`, so it's in none of the generated bindings and not in the SDK's `RelKind`. Also needs a representation decision, since members are deliberately stripped of `IN_COLLECTION` to keep them out of the scene explorer (ENG-8782).

## Testing

Rhino7, Rhino8 and Grasshopper8 all build clean (`-c Local`, 0 errors / 0 warnings), CSharpier-formatted. **Not yet exercised in Rhino** — no runtime verification of any of the six. Suggested order, cheapest to falsify first:

1. **ENG-9108** — render material on a layer, object on it with material source *by layer*; check viewer + Rhino receive.
2. **ENG-9111** — user text + user dictionary round trip; plus a Revit/IFC receive for parameters.
3. **ENG-9115 + ENG-9113** — receive a block-heavy model with multi-mesh elements **twice**: one generated definition set left, elements selecting as units from the report.
4. **ENG-9112** — one perspective and one parallel named view. The parallel case is what v1 got wrong and the ortho frustum rebuild is the least-exercised code here.
5. **ENG-9114** — one instance red, siblings default; also AutoCAD→Rhino.

### DIFF
--- a/Connectors/Rhino/Speckle.Connectors.RhinoShared/Operations/Receive/RhinoArtefactUserStrings.cs
+++ b/Connectors/Rhino/Speckle.Connectors.RhinoShared/Operations/Receive/RhinoArtefactUserStrings.cs
@@ -1,0 +1,110 @@
+using System.Globalization;
+using Rhino.DocObjects;
+
+namespace Speckle.Connectors.Rhino.Operations.Receive;
+
+/// <summary>
+/// Writes an artefact object's eav properties onto Rhino object attributes as user strings, so Rhino user text,
+/// nested user dictionaries and foreign parameters (Revit/IFC) survive an artefact receive [ENG-9111].
+/// </summary>
+/// <remarks>
+/// The key shape is the legacy one (<c>SpeckleAttributeExtensions.GetAttributes</c>): nested dictionaries flatten to
+/// dot-delimited paths, so <c>{ "User Dictionary": { "phase": 2 } }</c> becomes <c>User Dictionary.phase</c>. Values
+/// are formatted with the invariant culture — a Rhino user string is plain text, and formatting a double under the
+/// operator's locale would make the same bundle bake differently on different machines (and not re-parse on re-send).
+/// </remarks>
+internal static class RhinoArtefactUserStrings
+{
+  private const string PROPERTY_PATH_DELIMITER = ".";
+
+  /// <summary>
+  /// Bundle plumbing rather than user data: the root scalars every object carries
+  /// (<c>RhinoArtifactRootObjectBuilder.RootScalars</c>) and the hatch styling the send side stashes for
+  /// <see cref="RhinoHatchStyler"/>. All of it is either already applied to the native object (name, units, hatch
+  /// pattern) or reconstructed on the next send, so writing it as user text would only add noise the user then
+  /// sees in the Rhino properties panel.
+  /// </summary>
+  private static readonly HashSet<string> s_suppressedKeys = new(StringComparer.Ordinal)
+  {
+    "speckle_type",
+    "name",
+    "units",
+    "type",
+    "internalDefinitionName",
+    "hatchPatternName",
+    "hatchRotation",
+    "hatchScale",
+  };
+
+  /// <summary>Legacy noise filter: the trailing segment of a nested key, e.g. <c>area.units</c>, <c>area.name</c>.</summary>
+  private static readonly string[] s_suppressedSuffixes =
+  {
+    PROPERTY_PATH_DELIMITER + "units",
+    PROPERTY_PATH_DELIMITER + "name",
+    PROPERTY_PATH_DELIMITER + "internalDefinitionName",
+  };
+
+  /// <summary>
+  /// Flattens <paramref name="properties"/> onto <paramref name="atts"/> as user strings. A null or empty dictionary
+  /// is a no-op, so an object with no eav row costs nothing.
+  /// </summary>
+  public static void Apply(ObjectAttributes atts, Dictionary<string, object?>? properties)
+  {
+    if (properties is null || properties.Count == 0)
+    {
+      return;
+    }
+    Flatten(atts, properties, "");
+  }
+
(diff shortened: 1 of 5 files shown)


bimgeek (approved):
approved. there is a gap which will be tackled in a separate ticket.

PR #1502, opened 2026-08-05
fix(rhino): block members keep their layers on artefact round trip (ENG-9110)

kekesidavid:
Closes ENG-9110.

Objects inside a Rhino block definition came back on the base layer. After receive + explode, every member had lost its layer.

### Cause

ENG-8782 suppressed a definition member's top-level `DISPLAY` / `SOLID` / `DISPLAY_INSTANCE` **and** its `IN_COLLECTION`. Only the render edges caused the bug it was fixing (definition geometry drawing untransformed at the model origin) — but dropping the membership left the member's layer with nowhere to travel, so it never reached the bundle.

### Approach — SketchUp's, no spec change

`docs/artifact-object-model.md` proposed a new geometry-sourced `ON_LAYER` relation for this. It isn't needed, and the SketchUp connector had already solved the identical problem for tags (specklesystems/speckle-sketchup ENG-8851, `6df45ed`) without touching the spec:

- A member keeps a **carrier object row** — the same object-sourced `IN_COLLECTION` every top-level object emits.
- An **eav stamp** supplies the one direction nothing else can: from the K its definition reaches it by, back to that row. A member has no `DISPLAY` edge, so `ArtefactRelations.ObjectByGeometry()` cannot invert it.

**Send.** Restore `IN_COLLECTION` for members, then stamp `@speckle.geometry_k` with **every** geometry K the member owns — receive picks the authoritative 3dm solid *or* its display mesh(es) per member (`GroupDefinesByMember`), so whichever it picks has to resolve. A member that is itself a nested block owns no geometry, so it gets `@speckle.instance_k` with its INSTANCE node K instead.

**Receive.** `BuildDefinitions` resolves each member through its stamp to the carrier row, then through the **same** `ResolveLayer` every top-level object uses — shared `layerCache`, so a layer used by both a member and a scene object is created once — and sets it on the member's `ObjectAttributes`. Exploding a received instance now puts each member back on its source layer.

Keys and join are SketchUp's verbatim, so one reader serves both connectors. The only divergence: a Rhino member can own several geometries where a SketchUp one owns exactly one, hence the comma-joined list (a single value parses identically, so SketchUp bundles read here unchanged).

### Why the join lives in `Speckle.Connectors.Common`

`DefinitionMemberStamps` is host-agnostic: the keys are a cross-connector contract with SketchUp, AutoCAD blocks have the same gap waiting, and it's the only placement that makes the logic testable in this repo.

### The load-bearing invariant

Members stay out of the scene explorer because a carrier has **no render edge**, and every consumer that walks objects skips those — the Rhino native builder gates on `DISPLAY`/`SOLID`/`DISPLAY_INSTANCE`, and `ObjectsArtifactReader` drops an object whose geometry build returns null (both verified). Nothing in the spec or validator enforces this; it's called out in `DefinitionMemberStamps` and in both docs. Same exposure SketchUp has carried since 28 Jul, so if it turns out the viewer *does* show carriers, it's one fleet-wide fix rather than a Rhino-specific design flaw.

### Compatibility

Unstamped input — a pre-ENG-9110 bundle, or a member whose geometry didn't encode — falls back to the base layer, exactly the current behaviour.

### Tests

- **7 NUnit tests** (`Sdk/Speckle.Connectors.Common.Tests/Instances/`): multi-K fan-out to one owner, empty-geometry member, nested-instance join, SketchUp's bare-numeric form, unstamped bundle, malformed fragments skipped rather than thrown, and two members with a deliberate object-K/geometry-K numeric overlap.
- **SDK round-trip** pinning the carrier shape through parquet: specklesystems/speckle-sharp-sdk#530 (test-only, independent — this PR needs no SDK change).
- `Speckle.Rhino.slnx -c Local` builds clean; csharpier clean.

There is no Rhino test project, so the host-side bake is covered by manual round trips. On receive the session log (`%TEMP%\Speckle\sessions\*.summary.txt`) reports a new `memberStamps` stat — zero on a model with block members means the join was lost.

### Known gaps, documented not fixed

- **AutoCAD and SketchUp parity.** AutoCAD blocks have the same gap and today pin only the resolved member *colour* (ENG-8825); the mirror of this change is small and now shares the helper.
- **Member properties** are reachable as a side effect — the stamp is a genuine geometry→object route, which is what previously capped ENG-9111 for members. Nothing here consumes it.

### DIFF
--- a/Connectors/Rhino/Speckle.Connectors.RhinoShared/Operations/Receive/RhinoHostObjectArtefactBuilder.cs
+++ b/Connectors/Rhino/Speckle.Connectors.RhinoShared/Operations/Receive/RhinoHostObjectArtefactBuilder.cs
@@ -8,6 +8,7 @@ using Rhino.Render;
 using Speckle.Connectors.Common.Builders;
 using Speckle.Connectors.Common.Conversion;
 using Speckle.Connectors.Common.Diagnostics;
+using Speckle.Connectors.Common.Instances;
 using Speckle.Connectors.Common.Threading;
 using Speckle.Connectors.Rhino.Extensions;
 using Speckle.Connectors.Rhino.HostApp;
@@ -116,6 +117,11 @@ public class RhinoHostObjectArtefactBuilder : IArtifactHostObjectBuilder
     var doc = _converterSettings.Current.Document;
     var rels = bundle.Relations;
     var objByGeom = rels.ObjectByGeometry();
+    // ObjectByGeometry only covers DISPLAY edges, so it can't reach a block-definition member (a member has no
+    // render edge by design). The member stamps invert that missing direction — geometry / INSTANCE K → the member's
+    // object row — which is what BuildDefinitions needs to recover each member's layer [ENG-9110].
+    var memberIndex = DefinitionMemberStamps.Read(bundle.Properties);
+    session.SetStat("memberStamps", memberIndex.ObjectByGeometry.Count + memberIndex.ObjectByInstance.Count);
     var bakedObjectIds = new HashSet<string>();
     var conversionResults = new HashSet<ReceiveConversionResult>();
     // objK → its baked Rhino guids, tracked only for grouped objects (IN_GROUP) so step 5 can rebuild native groups.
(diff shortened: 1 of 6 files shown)


bimgeek (approved):
tested. it works as expected. conflicts should be resolved.

oguzhankoral (approved):

PR #1504, opened 2026-08-06
fix(autocad): three artefact round-trip gaps (ENG-9116, ENG-9117, ENG-9118)

kekesidavid:
Three AutoCAD artefact publish/receive gaps from the Speckle 7.0 backlog. Each is its own commit.

## ENG-9116 — block placement from a custom UCS

Publishing with a non-identity UCS misaligned block instances against the loose geometry they were drawn with, and transformed geometry inside block definitions twice.

The converter rebases every point/vector through `ReferencePointConverter`. The v1 builder pairs that with two compensations the artefact builder never carried over, both restored here:

- Definition members convert under a scoped `ReferencePointTransform = null` push, so they stay definition-local instead of being rebased a second time on top of their instance's transform. (`UnpackSelection`'s `atomicDefinitionObjectIds` is no longer discarded.)
- A top-level placement's transform is pre-multiplied by the inverse UCS, putting the world-authored transform back in the same frame as everything around it. Nested placements are already definition-local and untouched.

Identity-UCS documents are unaffected — `ReferencePointTransform` is null there and both branches collapse to the previous behaviour.

**Not covered:** the acceptance criterion asking for a unit test on the transform composition. It is `Matrix3d.PreMultiplyBy` + `TransformHelper.ConvertToInstanceMatrix4x4`, both AutoCAD API, and there is no AutoCAD-referencing test project in the repo to host an API-free test of it.

## ENG-9118 — ByLayer render materials

A ByLayer object got no material at all in the bundle. The unpacker's object stage skips those by design and its layer stage was never run — the artefact builder passed an empty layer list.

Running the layer stage alone isn't enough: a layer-sourced proxy lists the LAYER id, and phase 2 resolves proxy members only through `geometryKsByObjectId` (keyed by OBJECT id), so the MATERIAL node would land with no edge. Phase 1 now also records object → the layer whose material it inherits, gated on the layer carrying a non-Global material (the same condition the unpacker's layer stage uses, so every recorded object resolves to a MATERIAL node). Phase 2 inverts that and expands the layer-sourced member onto each inheriting object's geometry. Mirrors the Rhino ENG-9108 fix.

Layer records must stay open across the `UnpackMaterials` call, so collection and unpacking share one transaction. Directly assigned entity materials take the unchanged path. Definition members inherit through the same edge. A proxy member resolving to no geometry is now logged rather than dropped silently.

## ENG-9117 — ACI and explicit ByBlock colour semantics

An AutoCAD→AutoCAD round trip flattened every ACI colour to truecolor — same pixels, but the `ColorMethod` and index were gone, which breaks CTB plot styles, standards checks and index-reading scripts. Explicit ByBlock entities lost their inheritance too.

Widening the COLOR node is a spec decision, so this takes the AutoCAD-specific route the ticket describes: send writes the colour method (+ ACI index) as object properties alongside the unchanged ARGB edge; receive prefers them when rebuilding entity colour. Viewer rendering and non-AutoCAD consumers are unaffected — same `HAS_COLOR` edge, two extra eav scalars they ignore. Bundles without the keys (other hosts, older sends) take the existing ARGB path.

**Scope:** entities and block references, i.e. everything owning its colour as an object. Block-**definition members** are out of reach this way — their geometry carries no back-reference to an object in the bundle (which is why ENG-8825 resolved member layer colours at send time), so an ACI-coloured member still round-trips as truecolor. Members with no colour edge keep arriving ByBlock, unchanged.

## ENG-9119 — not fixed, documented

`docs/eng-9119-block-reference-materials-skipped.md`. The ticket's precondition fails: `HasMaterial` has no `srcIsObject` tag and there's no `MaterialByObject` on the read side, so an `InstanceProxy` (which owns no geometry) has nothing to hang the edge on. Needs a bundle-spec decision + SDK work first; the connector edits are ~3 lines after that. Rhino's ENG-9109 is blocked on the same tag. Analysis also posted on the ticket.

## Testing

Compile-verified only — **none of this has been run in AutoCAD yet.**

Heads-up for whoever builds this: `Local` does not build on `big-truck` as it stands, independently of this PR. All seven connectors call `new ObjectsArtifactPipeline(..., producedBy: slug)` but the SDK's ctor takes `ISpeckleApplication producer` (SDK `big-truck` @ `1d2201c`; `producedBy` only exists on `origin/bjorn/eng-9101-…`). To verify these changes I temporarily patched that one line, built AutoCAD 2026 / Civil3D 2026 / Plant3D 2026 clean (0 errors, 0 warnings), then reverted it rather than folding an unrelated repo-wide API fix into these commits.

Manual checks worth doing:
- rotated UCS (`UCS → Z → 45`), a line + a block touching its endpoint → publish → view and receive; they should stay coincident, and definition geometry should not be doubly transformed
- material on a layer, a `Solid3d` on it left ByLayer → publish → view and receive
- one line ACI 1, one explicitly ByBlock → publish → receive → inspect `ColorMethod` and index
- a plain identity-UCS send with truecolor objects, to confirm nothing regressed

### DIFF
--- a/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadColorSemanticKeys.cs
+++ b/Connectors/Autocad/Speckle.Connectors.AutocadShared/HostApp/AutocadColorSemanticKeys.cs
@@ -1,0 +1,17 @@
+namespace Speckle.Connectors.Autocad.HostApp;
+
+/// <summary>
+/// Object-property keys carrying the AutoCAD colour semantics an artefact COLOR node cannot express — it stores a
+/// plain ARGB, so an AutoCAD→AutoCAD round trip flattened every ACI colour to truecolor and every explicit ByBlock
+/// to a fixed value [ENG-9117]. The send builder writes these alongside the ARGB edge (so the viewer and non-AutoCAD
+/// consumers are unaffected); the receive builder prefers them when rebuilding the entity colour.
+/// </summary>
+internal static class AutocadColorSemanticKeys
+{
+  /// <summary><c>"aci"</c> (an AutoCAD Color Index colour, index in <see cref="Index"/>) or <c>"block"</c> (the
+  /// entity was explicitly set to ByBlock). Absent for a plain truecolor object, whose ARGB is already lossless.</summary>
+  public const string SOURCE = "autocadColorSource";
+
+  /// <summary>The original AutoCAD Color Index (0..256), present only when <see cref="SOURCE"/> is <c>"aci"</c>.</summary>
+  public const string INDEX = "autocadColorIndex";
+}
--- a/Connectors/Autocad/Speckle.Connectors.AutocadShared/Operations/Receive/AutocadHostObjectArtefactBuilder.cs
+++ b/Connectors/Autocad/Speckle.Connectors.AutocadShared/Operations/Receive/AutocadHostObjectArtefactBuilder.cs
@@ -206,6 +206,8 @@ public class AutocadHostObjectArtefactBuilder : IArtifactHostObjectBuilder
           }
           materialIdByObject.TryGetValue(appId, out ObjectId objMaterial);
           bool hasObjColor = colorArgbByObject.TryGetValue(appId, out int objArgb);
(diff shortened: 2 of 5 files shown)


bimgeek (approved):
tested all tickets. confirmed the fix.

PR #1506, opened 2026-08-06
fix(rhino): fix per-mesh materials, type parameters, and instance materials on artefact receive

bjoernsteinhagen:
## Description

Three related Rhino artefact round-trip bugs, fixed together for easier testing:

- **ENG-9153**: multi-material objects (e.g. a Revit wall) collapsed all their meshes onto one material on receive, instead of keeping each mesh's own.
- **ENG-9136**: Type Parameters / System Type Parameters never reached Rhino on receive — only instance parameters did.
- **ENG-9109**: a material assigned directly to a block instance (`MaterialFromObject` on the placement) was silently dropped on send.

Fixes [ENG-9153](https://linear.app/speckle/issue/ENG-9153/rhino-artifact-receive-loses-per-mesh-materials), Fixes [ENG-9136](https://linear.app/speckle/issue/ENG-9136/rhino-type-parameters-are-lost-on-artefact-receive), Fixes [ENG-9109](https://linear.app/speckle/issue/ENG-9109/rhino-block-instance-materials-are-lost-on-artefact-round-trip)

Companion SDK PR: https://github.com/specklesystems/speckle-sharp-sdk/pull/534

### ENG-9153 — per-mesh materials
- [x] Source model with distinct material regions
- [x] Received in Rhino, each mesh shows its own material (not all the same)
- [x] Re-receiving 2–3 times gives the same assignment every time (order-independent)
- [x] A plain single-material object still receives its material correctly
- [x] A block member's material (geometry-sourced) still works inside its definition

image

### ENG-9136 — type parameters
- [x] Revit wall with a Type Parameter set, received in Rhino, shows `properties.Parameters.Type Parameters.*` in user text alongside `Instance Parameters.*`

image

### ENG-9109 — instance material
- [x] Block with two instances, one given a direct `MaterialFromObject` override, publishes with an instance-sourced `HAS_MATERIAL` edge
- [x] Web viewer renders the overridden instance differently from the untouched one
- [x] Received back into a fresh Rhino doc, the overridden instance shows `MaterialFromObject` with the correct material
- [x] The untouched instance still renders via the definition's own material

image

## Checklist:

### DIFF
--- a/Connectors/Rhino/Speckle.Connectors.RhinoShared/Operations/Receive/RhinoHostObjectArtefactBuilder.cs
+++ b/Connectors/Rhino/Speckle.Connectors.RhinoShared/Operations/Receive/RhinoHostObjectArtefactBuilder.cs
@@ -139,9 +139,10 @@ public class RhinoHostObjectArtefactBuilder : IArtifactHostObjectBuilder
     onOperationProgressed.Report(new("Converting materials", null));
     Dictionary<string, Guid> materialByObject;
     Dictionary<int, Guid> materialByGeometry;
+    Dictionary<int, Guid> materialByInstance;
     using (session.Phase("Materials"))
     {
-      (materialByObject, materialByGeometry) = CreateMaterials(doc, bundle, objByGeom);
+      (materialByObject, materialByGeometry, materialByInstance) = CreateMaterials(doc, bundle, objByGeom);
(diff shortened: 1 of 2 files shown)


bimgeek (approved):
tested. can confirm the fix.

PR #1519, opened 2026-08-12
feat(revit): export language-stable unit ids instead of localized unit labels

bimgeek:
## Description

The Revit connector exported unit labels through `LabelUtils.GetLabelForUnit`, which returns text in the running Revit session's display language. A German Revit publish writes `Quadratmeter` where an English one writes `Square meters`, so downstream tooling that matches on English labels (Automate carbon workflows, dashboards, material stats) fails or returns empty results for localized exports.

The `units` field now carries the unversioned Forge schema unit name instead, which is identical in every display language:

```jsonc
// before (German Revit)
{ "name": "area", "value": 42.5, "units": "Quadratmeter" }
// after (any Revit language)
{ "name": "area", "value": 42.5, "units": "squareMeters" }
```

This applies to material quantity properties (`area`, `volume`, `length`, `density`, `compressiveStrength`) and to all double-valued element parameters.

Two decisions worth explaining:

- Forge schema name rather than the unit symbol, because Revit localizes symbols in some locales too (Cyrillic `м²`).
- The version suffix in the raw Forge id (`autodesk.unit.unit:squareMeters-1.0.1`) is stripped, because it can change between Revit releases and would break downstream matching.

Side effect: double parameters without a valid unit type used to throw inside `GetLabelForUnit`, and the catch in `ParameterExtractor` then skipped the whole parameter. They are now exported without a `units` entry.

Resolves [ENG-8735](https://linear.app/speckle/issue/ENG-8735/revit-connector-should-export-language-stable-units-for-material)

## User Value

Users on non-English Revit installs get working carbon, dashboard, and material stats workflows, since unit metadata no longer depends on the Revit display language.

## Changes:

- New `ForgeTypeIdExtensions.GetStableUnitsId()`: extracts `squareMeters` from `autodesk.unit.unit:squareMeters-1.0.1`, returns null for empty ids.
- `MaterialQuantitiesToSpeckleLite.AddMaterialProperty` and `ParameterDefinitionHandler.HandleDefinition` use it instead of `LabelUtils.GetLabelForUnit`.

## Validation of changes:

- Tested manually on Windows: the same model published from English and German Revit produces identical `units` values for material quantities and dimension parameters.
- No unit tests added: `Speckle.Converters.RevitShared.Tests` is a shared project that no compiled test project imports, so tests there never run. The parsing logic lives in one small extension method.

## Checklist:

- [ ] I have added appropriate tests. (see Validation section)
- [ ] I have updated or added relevant documentation. (n/a)

### DIFF
--- a/Connectors/Rhino/Speckle.Connectors.RhinoShared/Operations/Receive/RhinoHostObjectArtefactBuilder.cs
+++ b/Connectors/Rhino/Speckle.Connectors.RhinoShared/Operations/Receive/RhinoHostObjectArtefactBuilder.cs
@@ -461,8 +461,7 @@ public class RhinoHostObjectArtefactBuilder : IArtifactHostObjectBuilder
   // doesn't carry, so a blob with either has to go through the full decoder instead [ENG-9214]. n-gons are safe here —
   // they ride the face array, and BuildMesh rebuilds their MeshNgon records.
   private static bool IsFastPathMesh(SgeoHeader header) =>
-    header.PrimitiveType == SgeoPrimitiveType.Mesh
-    && (header.Flags & (SgeoFlags.HasNormals | SgeoFlags.HasUvs)) == 0;
+    header.PrimitiveType == SgeoPrimitiveType.Mesh && (header.Flags & (SgeoFlags.HasNormals | SgeoFlags.HasUvs)) == 0;
 
   // SGEO neutral mesh → Rhino mesh (Speckle count-prefixed face format; matches MeshToHostConverter).
   private static RG.Mesh BuildMesh(SgeoMesh sm)
--- a/Converters/Revit/Speckle.Converters.RevitShared/Extensions/ForgeTypeIdExtensions.cs
+++ b/Converters/Revit/Speckle.Converters.RevitShared/Extensions/ForgeTypeIdExtensions.cs
@@ -23,4 +23,30 @@ public static class ForgeTypeIdExtensions
   {
     return forgeTypeId.TypeId;
   }
+
+  /// <summary>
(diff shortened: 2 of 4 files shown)


bimgeek:
Note for reviewers: the CI format check was failing on RhinoHostObjectArtefactBuilder.cs, which landed unformatted on big-truck via #1515 and is unrelated to this PR. Commit 8f39eeb7 formats it so the check can pass. Happy to drop that commit if you'd rather fix it on big-truck directly.

bimgeek:
Second base-branch issue: central package versions changed on big-truck without regenerating packages.lock.json files, so locked-mode restore failed with NU1004 for every project in the solution. Commit f15db40d refreshes the locks (81 files, verified locked-mode restore passes locally). Like the formatting commit, this is unrelated to the Revit change and can be split into its own PR if preferred.

bimgeek:
Update on CI: I removed the lock-file commit after digging further. The failing Test job is structural on big-truck, not fixable from this PR:

- big-truck's packages.lock.json files pin Speckle.Sdk/Speckle.Objects 1.0.0, which are locally-packed dev builds of speckle-sharp-sdk (they carry the new Speckle.Sdk.Pipelines.Receive / ArtefactBundle APIs). They are not on nuget.org, so CI locked-mode restore fails with NU1004.
- Regenerating the locks against the released 2026.6.0 lets restore pass but the build then fails, because big-truck code uses SDK APIs that 2026.6.0 does not have. It would also break local dev builds that rely on the 1.0.0 pins.

So CI stays red for every big-truck PR until a Speckle.Sdk with the new pipeline APIs is published (or CI gets a feed with dev SDK packages). This PR now contains only the Revit unit fix plus the csharpier fix for the file from #1515. The Revit change itself builds clean and was tested manually on English and German Revit.

PR #1527, opened 2026-08-18
fix(tekla): re-send objects after a coordinate frame change

bjoernsteinhagen:
## Description

Tekla reports geometry in the model's current coordinate frame (work plane + current base point). Changing the frame moves every object's coordinates without changing any object, so `ModelObjectChanged` never fires and nothing clears the send cache — publishes keep re-using objects converted in the old frame. The cache is keyed on object id + project, not model card, so a new card doesn't help.

Now we re-read the frame before each send and clear the cache when it moved. Grids were a separate pre-existing bug found while testing: they ignored the frame entirely.

Fixes ENG-8768

## User Value

Publishing after changing true north / the coordinate system sends the objects as they actually are, instead of silently re-publishing the old orientation.

## Changes:

- new `TeklaCoordinateSystemTracker`: fingerprints work plane, north direction and current base point, clears the send cache on change. Unknown or unreadable frame also clears, so it fails safe. This is what Revit and AutoCAD does
- `TeklaSendBinding.Send` calls it before objects are gathered and converted
- grids placed through their own coordinate system so they follow the frame. Also fixes grids whose origin isn't 0,0, and a scale factor taken from `AxisX.X` that shrank with rotation (~1% oversize at -8°) and divided by zero at 90°

## Validation of changes:

- published once, cache cleared
- re-published immediately, cache used as base point hadn't changed
- created a new base point, republished, cache cleared

image

### DIFF
--- a/Connectors/Tekla/Speckle.Connector.TeklaShared/Bindings/TeklaSendBinding.cs
+++ b/Connectors/Tekla/Speckle.Connector.TeklaShared/Bindings/TeklaSendBinding.cs
@@ -10,6 +10,7 @@ using Speckle.Connectors.DUI.Models;
 using Speckle.Connectors.DUI.Models.Card;
 using Speckle.Connectors.DUI.Models.Card.SendFilter;
 using Speckle.Connectors.DUI.Settings;
+using Speckle.Connectors.TeklaShared.HostApp;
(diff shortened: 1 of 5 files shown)


bimgeek (approved):
YOLO

PR #1529, opened 2026-08-18
fix(tekla): publish tekla's object type instead of the api class

bjoernsteinhagen:
## Description

We published the .NET class name as `type`, and Tekla's classes are much coarser than its object types — beams, columns, panels and both footing kinds are all `Beam`, plates and slabs are all `ContourPlate`. So most of a model came through labelled "Beam".

Now we publish the type Tekla itself reports (`Beam.Type`, `PolyBeam.Type`, `ContourPlate.Type`), normalised to PascalCase, and fall back to the class name for objects that don't carry one (bolts, rebar, grids).

Normalising keeps real beams on `"Beam"`, so only the objects that were mislabelled change value. Plates and slabs do move off `"ContourPlate"` — worth a release note for anyone filtering on `type` downstream.

## User Value

Better differentiation as to what objects are. Not perfect, but better. Constrained by API.

## Changes:

- new `GetSpeckleType()` in the converter, used by the `type` field, the collection tree and the selection filter summary, so the three can't drift apart
- `ContourPlate` reporting `UNKNOWN` keeps the class name rather than publishing "Unknown"
- API class kept in properties as `apiClass`, so a `PolyBeam` publishing as `"Beam"` is still identifiable. Not named `class` to avoid colliding with Tekla's numeric CLASS report property

### Before

image

### After

image

### DIFF
--- a/Connectors/Tekla/Speckle.Connector.TeklaShared/Bindings/TeklaSelectionBinding.cs
+++ b/Connectors/Tekla/Speckle.Connector.TeklaShared/Bindings/TeklaSelectionBinding.cs
@@ -1,5 +1,6 @@
 using Speckle.Connectors.DUI.Bindings;
 using Speckle.Connectors.DUI.Bridge;
+using Speckle.Converters.TeklaShared.Extensions;
 using Tekla.Structures;
 using Tekla.Structures.Model;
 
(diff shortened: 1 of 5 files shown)


bimgeek (approved):
YOLO

PR #1543, opened 2026-08-24
fix(revit): further qa/qc fixes

bjoernsteinhagen:
## Description

Four Revit regressions on `big-truck`.
- [ENG-9338: Revit: wall, floor and roof layer structure is missing from published data](https://linear.app/speckle/issue/ENG-9338/revit-wall-floor-and-roof-layer-structure-is-missing-from-published)
- [ENG-9337: Revit: received elements land in the wrong category when host and source Revit languages differ](https://linear.app/speckle/issue/ENG-9337/revit-received-elements-land-in-the-wrong-category-when-host-and)
- [ENG-8684: Type Name eav prop for revit files seems to have slipped post normalisation](https://linear.app/speckle/issue/ENG-8684/type-name-eav-prop-for-revit-files-seems-to-have-slipped-post)
- [ENG-9263: Revit: linked-model transform hash is lossy — coincident-origin placements collide into one identity](https://linear.app/speckle/issue/ENG-9263/revit-linked-model-transform-hash-is-lossy-coincident-origin)

## Changes:

- `CompoundStructureExtractor`: layer buildup at `properties.Structure`, ordinal-keyed, material-less layers kept
- `RevitHostObjectArtefactBuilder`: read nested `builtInCategory`; identifier → display name → Generic Models as a real ladder; cache key from both values
- `LinkedModelHandler`: placement identity from `RevitLinkInstance.UniqueId`, for the container key and the app-id suffix alike, on the 4.0 and legacy paths. `GetTransformHash` and `FindLinkInstanceForDocument` deleted
- `ParameterExtractor`: `Type Name` guaranteed; caches keyed by UniqueId rather than `ElementId` (unique only *within* a document, while the extractor spans host + every link)

### ENG-9338
Model with a multi-layer wall, floor, roof, ceiling; one layer with no material; one curtain wall:
- [x] `Structure` keyed `0,1,2…`, each with `material` / `function` / `thickness` + unit
- [x] Order matches Revit's Edit Assembly dialog (exterior → interior)
- [x] Thickness in document display units

image

### ENG-9337
Model with walls, doors, windows, families:
- [x] Load into German Revit → source categories, not Generic Models
- [x] Both with "receive instances as families" ON and OFF

image

### ENG-9263
Same RVT linked twice at the same origin, +90°/−90°:
- [x] Two separate Model nodes; both placements' geometry present
- [x] Rotational array → one occurrence per placement

Host-model:
- [x] Reference point set to Survey / Shared Coordinates → editing a **host** wall parameter succeeds (no "Cannot modify elements from a linked model")

### ENG-8684
- [x] `Type Name` under Type Parameters → Identity Data = the type's actual name

## Checklist:

### DIFF
--- a/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
+++ b/Connectors/Revit/Speckle.Connectors.RevitShared/Bindings/RevitSendBinding.cs
@@ -306,16 +306,16 @@ internal sealed class RevitSendBinding : RevitBaseBinding, ISendBinding
             document,
             modelCard.SendFilter,
             linkedDoc,
-            transform
+            linkedModel
           );
-          linkedDocumentContexts.Add(new(transform, linkedDoc, linkedElements));
+          linkedDocumentContexts.Add(new(transform, linkedDoc, linkedElements, linkedModel));
         }
         // ⚠️ when disabled, still adds empty contexts to maintain warning generation in RevitRootObjectBuilder
         // this approach (to signal that warnings are needed) relies on empty element lists which smells and is a bit of an implicit mechanism
         // buuuuut, it works (for now 👀).
         else
         {
-          linkedDocumentContexts.Add(new(transform, linkedDoc, new List<Element>()));
+          linkedDocumentContexts.Add(new(transform, linkedDoc, new List<Element>(), linkedModel));
         }
       }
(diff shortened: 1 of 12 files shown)


bimgeek (approved):
If you've tested enough, we can merge.

PR #1554, opened 2026-08-27
fix(revit): restore v1 grouping on artefact receive, free the Comments parameter (ENG-8805)

kekesidavid:
Closes [ENG-8805](https://linear.app/speckle/issue/ENG-8805/no-groups-comments-parameter-hijacked-for-tracking-duplicates-on).

## What was wrong

The artefact receive path organised nothing and tracked its bake by stamping every element's **Comments** parameter with `Project X: Model Y`. Comments is user-visible, schedulable and commonly used for documentation — and because the same field was the tracking key, a user editing it orphaned that element into a **duplicate** on the next receive.

Since the ticket was filed, the upgrade-path duplication (regression #3) was already fixed in both directions. What remained: no grouping (#1), the Comments hijack (#2), and rename orphans (#4). Note the ticket's scope line is stale — the artefact builder is no longer `#if NET8_0_OR_GREATER`-gated, so this affected **Revit 2023–2027**, not just 2025+.

## What this does

**Grouping is back, v1-style.** Everything baked lands in one pinned, named top-level `Group` — the user's handle on a received model (one click to select/move/delete) and the receive's tracking key. Baked in its own transaction after painting, matching v1's bake → paint → group order, because an element's faces only exist once the bake commits and painting a group member risks group-inconsistency warnings. If Revit refuses the group (already-grouped elements, differing design options/worksets) the receive logs and keeps the geometry rather than failing.

**Comments is free.** What survives of the marker goes into a hidden per-model manifest — Extensible Storage on a `DataStorage` element — recording the Group and the Materials the receive created, keyed on project/model **ids** so a rename on the web no longer strands the prior bake.

**No send-side change was needed.** The marker was already derived from names the receive operation passes; the ids were already on `ReceiveInfo`, just not forwarded to the builder. `IArtifactHostObjectBuilder.Build` now takes an `ArtefactReceiveTarget` (ids + names) — mechanical update for Rhino and AutoCAD, which use the names as before.

## The PurgeMaterials finding

Worth reading on its own, because "port `PurgeMaterials` to the artefact path" would have been the wrong fix.

`RevitMaterialBaker.PurgeMaterials(baseGroupName)` deleted every `Material` whose name *contained* the sanitised base group name. But `BakeMaterial` names materials after the Speckle material alone, with **no project/model suffix** — so the predicate never matched anything it created. It was a **no-op on `dev` as well**, and what actually prevents material accumulation in v1 is reuse-by-name in `FindExistingMaterialByName`. It also carried an unscoped delete: a user's own material named after the project or model would have been removed.

It now deletes exactly the materials recorded in the manifest, and only ones the receive **created** — never ones it reused from the document, which are the user's. This also closes the artefact path's real gap: when a material's colour changed between versions a second `Foo AARRGGBB` element was minted and the original orphaned.

## Cleanup order

A document can hold a bake from any generation of this connector, so `PurgePriorReceive` runs newest mechanism first: manifest Group by UniqueId (rename-proof) → manifest Materials → name-based `PurgeGroups` (v1 and pre-manifest bakes) → a **transitional** Comments sweep for bakes made between the direct-bake path landing and this change. That last one is marked removable once no such document is in circulation.

Grouping and tracking are now one concern in `RevitReceiveTracker`, used by **both** receive paths, so a document received into by either is cleaned up correctly by the other.

## Known limitation

Inherited from v1, unchanged: explicitly ungrouping a received model leaves its elements untracked, and the next receive bakes alongside them. Tracking every element individually would mean a multi-megabyte manifest on a large model for no gain over the group purge.

## Testing

Builds clean (`Local.slnx -c Local`, all Revit versions + Rhino/AutoCAD/Civil3D). There is no test harness for the Revit host builders, so this needs manual QA:

| # | Scenario | Expected |
|---|---|---|
| a | Receive twice into the same doc | one group, no duplicate geometry |
| b | Edit Comments on a baked element between receives | no duplicate — tracking no longer keys on Comments |
| c | Bake on a pre-change `big-truck` build, then re-receive on this one | transitional sweep cleans it |
| d | v1 bake → artefact receive, and artefact bake → v1 receive | both cleaned |
| e | **Rename the model on web, re-receive** | prior bake cleaned (this is the new one) |
| f | Large model | check the `Grouping` phase duration in the artefact session log — this is the perf cost that motivated dropping groups |
| g | `Receive Blocks as Families` on **and** off | both bake sites join the group |
| h | Camera views | still purged, never grouped |
| i | Receive, change a material's colour upstream, re-receive | old material cleaned, no orphan |

### DIFF
--- a/Connectors/Autocad/Speckle.Connectors.AutocadShared/Operations/Receive/AutocadHostObjectArtefactBuilder.cs
+++ b/Connectors/Autocad/Speckle.Connectors.AutocadShared/Operations/Receive/AutocadHostObjectArtefactBuilder.cs
@@ -102,12 +102,13 @@ public class AutocadHostObjectArtefactBuilder : IArtifactHostObjectBuilder
 
   public Task<HostObjectBuilderResult> Build(
     ArtefactBundle bundle,
-    string projectName,
-    string modelName,
+    ArtefactReceiveTarget target,
     IProgress<CardProgress> onOperationProgressed,
     CancellationToken cancellationToken
   ) =>
-    _threadContext.RunOnMain(() => BakeAll(bundle, projectName, modelName, onOperationProgressed, cancellationToken));
+    _threadContext.RunOnMain(() =>
+      BakeAll(bundle, target.ProjectName, target.ModelName, onOperationProgressed, cancellationToken)
+    );
 
 #pragma warning disable CA1506
   private HostObjectBuilderResult BakeAll(
--- a/Connectors/Revit/Speckle.Connectors.RevitShared/DependencyInjection/RevitConnectorModule.cs
+++ b/Connectors/Revit/Speckle.Connectors.RevitShared/DependencyInjection/RevitConnectorModule.cs
@@ -105,6 +105,8 @@ public static class ServiceRegistration
     serviceCollection.AddScoped<RevitFamilyBaker>();
     serviceCollection.AddScoped<FamilyGeometryBaker>();
     serviceCollection.AddScoped<RevitGroupBaker>();
+    serviceCollection.AddScoped<RevitReceiveManifest>();
+    serviceCollection.AddScoped<RevitReceiveTracker>();
     serviceCollection.AddScoped<RevitMaterialBaker>();
     serviceCollection.AddScoped<RevitViewBaker>();
     serviceCollection.AddScoped<RevitViewManager>();
--- a/Connectors/Revit/Speckle.Connectors.RevitShared/HostApp/RevitGroupBaker.cs
+++ b/Connectors/Revit/Speckle.Connectors.RevitShared/HostApp/RevitGroupBaker.cs
@@ -23,17 +23,40 @@ public class RevitGroupBaker : TraversalContextUnpacker
 
   public void AddToTopLevelGroup(Element revitElement) => _elementIdsForTopLevelGroup.Add(revitElement.Id);
 
-  public void BakeGroupForTopLevel(string baseGroupName)
+  public string? BakeGroupForTopLevel(string baseGroupName) =>
+    BakeGroupForTopLevel(baseGroupName, _elementIdsForTopLevelGroup);
+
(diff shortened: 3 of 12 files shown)


kekesidavid:
### Follow-up: materials are reused as-is again (6a1d841)

Confirmed from the code, and the cause was two layers deep.

**`dev`** — `RevitMaterialBaker.BakeMaterial` calls `FindExistingMaterialByName`, and on a hit `return`s that id immediately. It never reads or writes the existing material's properties, so a user's recolour survives.

**plain `big-truck`** — `FindOrCreateMaterial` looked up `"Roof"`, found it yellow, compared against the incoming red via `SameColor`, and on the mismatch retargeted to `"Roof FFFF0000"`. That didn't exist, so it created a **second, red material** and painted the roof with it. The user's yellow `"Roof"` stayed in the document as an orphan nothing referenced.

**this branch, before 6a1d841** — worse: the material tracking I added deleted the recorded `"Roof"` in pre-clean first, so the edited material was destroyed outright and recreated red.

Now aligned with `dev`: an existing same-named material is reused as-is, full stop. The trade-off comes along with it — two source materials that share a name but not a colour collapse onto whichever one the document already holds. That is `dev`'s behaviour, which is what was asked for.

**Knock-on: the material purge is gone entirely.** Reuse-by-name cannot accumulate duplicates across re-receives, so there is nothing left to clean up — the created-material tracking and the manifest-driven purge both came out, along with the `SameColor` helper. The manifest now records only the group.

The finding about the old `PurgeMaterials(baseGroupName)` still stands (it never matched anything `BakeMaterial` creates, and carried an unscoped delete), but it is now simply **deleted rather than replaced**, with a note in `RevitMaterialBaker` recording why there is deliberately no purge. Its schema slot is kept, written empty, so the shape still matches manifests written by the build already tested.

Full `Local.slnx -c Local` build is green. Worth re-running QA row (i) — receive, recolour a material, re-receive — plus a quick check that (a) and (e) still behave, since pre-clean changed.

bimgeek (approved):

PR #1568, opened 2026-09-07
feat(send): report the connector version on ingestion source data

danielgak:
> **Draft — blocked on an SDK release.** `SourceDataInput.connectorVersion` lands in speckle-sharp-sdk #575; this branch does not compile against the pinned `Speckle.Sdk 2026.9.0-beta.4`. Bump the pin in `Directory.Packages.props` once a package with https://github.com/specklesystems/speckle-sharp-sdk/pull/575 is out, then un-draft.

## What

The three `Ingestion.Create` call sites in `SendOperation.cs` pass `speckleApplication.SpeckleVersion` as the new optional `connectorVersion`, next to the host-app release they already send as `sourceApplicationVersion`.

`SpeckleVersion` is the connector's own build here: `Connector.Initialize` registers `Speckle.Connectors.Common`'s assembly version as the SDK's `speckleVersion`, which is also what `GetConnectorVersion()` exposes to DUI.

## Why

Server-side (speckle-server-internal #2815) the Ingestion Details modal shows a **Connector version** row when the field is present. Until now the connector build was nowhere on the wire; support could only see "Revit 2024".

## Compatibility

A set `connectorVersion` is rejected by servers that predate #2815 (unknown input field). app.speckle.systems tracks main so it has the field before any connector release ships; self-hosted servers need the release that includes it. Callers that leave it null are unaffected (nulls are not serialized).

https://claude.ai/code/session_01HUhU8RM42XgXbWLXeLgD7D

### DIFF
--- a/Connectors/Revit/Speckle.Connectors.RevitShared.Cef/Plugin/CefSharpPanel.xaml.cs
+++ b/Connectors/Revit/Speckle.Connectors.RevitShared.Cef/Plugin/CefSharpPanel.xaml.cs
@@ -2,8 +2,8 @@ using System.Windows.Controls;
 using System.Windows.Threading;
 using Autodesk.Revit.UI;
 using CefSharp;
+using Speckle.Connectors.Common.Settings;
 using Speckle.Connectors.DUI.Bridge;
-using Speckle.Connectors.DUI.Settings;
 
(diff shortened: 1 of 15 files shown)


danielgak:
can you take this cc @JR-Morgan ?

JR-Morgan (approved):

PR #1580, opened 2026-09-16
fix(grasshopper): version message on ingestion-created version

bjoernsteinhagen:
## Validation of changes:

- Tested both publish sync and async

image


jsdbroughton (approved):
great!

